PR #26, opened 2016-10-03
[MB-1919] update to iOS SDK 8.0.2

marc-scig:
I think this is all of it...

Edit: updated to 8.0.2


src/AirshipBindings.iOS/ApiDefinition.cs
```diff
@@ -269,8 +387,256 @@ interface UAirship
...
+		[NullAllowed, Export("localizationKeys", ArgumentSemantic.Copy)]
+		NSDictionary LocalizationKeys { get; }
+
+		// @property (readonly, copy, nonatomic) NSDictionary * _Nonnull notificationInfo;
+		[Export("notificationInfo", ArgumentSemantic.Copy)]
+		NSDictionary NotificationInfo { get; }
+
+		// @property (readonly, nonatomic, strong) UNNotification * _Nullable notification;
+		[NullAllowed, Export("notification", ArgumentSemantic.Strong)]
+		UNNotification UNNotification { get; }
```
marc-scig: note: this property is simply "notification" in the original interface, but C# will not allow us to have that and the factory constructors below.

rlepinski:
This looks good... its hard to review but I think you got all the things.

PR #27, opened 2016-10-05
Fix android build

rlepinski:
Turns out we can restore a solution to download missing nuget packages. Seems to correct the missing bindings issue.


hcrowell:
🍊

PR #28, opened 2016-10-05
Use carthage to install AirshipKit

rlepinski:
Before opening the solution you need to install AirshipKit with `carthage update`


marc-scig:
Should we call out the carthage-ness in the README?

PR #30, opened 2016-10-07
[MB-2009] AirshipExtensionBindings

marc-scig:
This gets us closer to full extension support. What we get:

* A separate bindings project for AirshipAppExtensions.framework (carthage based)
* A nuspec for generating a separate nuget package for those bindings
* build script updates

What we don't

* A service extension sample in the component. This doesn't appear to be possible at the moment, due to the magic xamarin-component.exe does and its inability to recognize the project type

However, creating an extension for an existing app and linking against the package should be pretty easy.


rlepinski:
👍

marc-scig:
Oh yeah, right-o

PR #33, opened 2016-10-12
Combine airship bindings

rlepinski:
Moves the extension bindings into the airship bindings. Only way I could get the framework to actually copy into the app bundle.


hcrowell:
🍊

marc-scig:
let's get the android version numbers updated as well. I can push that

rlepinski:
@marc-scig Android is versioned using the shared assembly. I believe we are good to go.

marc-scig:
Oh right.

PR #34, opened 2016-10-13
Fix iOS notification service extension

rlepinski:
Fixes the extension for reals this time. 

The main airship bindings needs to contain the AirshipExtensions framework so its actually available in the app bundle. Extensions dependencies are ignored (Xamarin limitation). However we cant have the extension reference the main framework, for whatever reasons that fails. So we have the main package that includes both frameworks and an extension only framework for the extension. Yay.


marc-scig:
🎱  🚋  🚜 🇧🇲

PR #48, opened 2017-02-14
[MB-2099] Generate documentation & update build process for PCL.

madelgi:
[MB-2099] Generate documentation for PCL.


docs/build.gradle
```diff
@@ -0,0 +1,16 @@
+task clean(type: Delete) {
```
rlepinski: Do we want to use gradle here? can we do it with just an .sh file?

rlepinski:
👍 

I am cool with using gradle but it would be nice to not have to use bash and gradle

docs/build.gradle
rlepinski: We should probably remove the gradle file for now

docs/mainpage.md
```diff
@@ -0,0 +1,20 @@
...
+
+
+## Requirements
+ - Xamarin.iOS and Xamarin.Android
+
+## Resources:
+ - [Getting started guide](http://docs.urbanairship.com/platform/xamarin.html)
+
+## Setup
+Use the Xamarin component store or NuGeT to install the urbanairship-portable
```
hcrowell: NuGeT -> NuGet* ?

hcrowell:
🍊  after reviewing comment above ^

PR #50, opened 2017-03-07
[MB-2131] Convert xamarin build process to gradle

madelgi:
Should make it automatically pull down most recent android sdk version for android bindings. Also can maybe move the update version functionality to gradle.


rlepinski:
👍 

Yes please move the update version stuff into the gradle build system. It would be nice if it could update the other versions based on the airship.properties file.

rlepinski:
👍  Assuming it works.

gradlew.bat
```diff
@@ -0,0 +1,84 @@
+@if "%DEBUG%" == "" @echo off
```
BrianBatchelder: Just have to ask - do we ever build on Windows?

rlepinski: nope, this is just the standard gradle wrapper

BrianBatchelder: Thought I'd seen it before. Guess it doesn't hurt, but its inclusion might kinda imply we tested it.

madelgi: i think the idea is to make it so folks on windows can build the repo on their systems as well, even if we'll never do it ourselves

BrianBatchelder: Ah. My original question should have been "does anyone ever build on Windows?".

Thanks, Max.

PR #51, opened 2017-03-10
Resolve Issue with urbanairship.portable install into PCL

bhazen:
Currently when attempting to install the urbanairship.portable package into a PCL in a Xamarin Forms project the install will fail because it depends on urbanairship which has no assembly to match that profile.

Updated the nuspec file so only platform specific version of urbanairship.portable have that dependency.

### DIFF
--- a/UrbanAirship.Portable.nuspec
+++ b/UrbanAirship.Portable.nuspec
@@ -9,7 +9,13 @@
       <description>A portable client library for the Urban Airship SDK.</description>
 
       <dependencies>
-          <dependency id="urbanairship" version="4.1.0"/>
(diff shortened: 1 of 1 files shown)


madelgi:
@bhazen :+1: This looks good to me -- before I merge this, would you mind signing our contribution agreement? http://docs.urbanairship.com/contribution-agreement.html

bhazen:
@madelgi signed

madelgi:
@bhazen This should be resolved in version 4.1.1 [here](https://www.nuget.org/packages/urbanairship.portable). Thanks again for bringing this to our attention!

PR #54, opened 2017-03-13
[MB-2138] Updated bindings and fixed registration delegate imp

crow:
- Updated the bindings with proper memory assignment.
- Fixed the registration delegate implementation.


samples/ios-unified/AppDelegate.cs
```diff
@@ -22,6 +22,7 @@ public class AppDelegate : UIApplicationDelegate
 	{
 		private InboxDelegate inboxDelegate;
 		private PushHandler pushHandler;
+		private RegistrationDelegate registrationDelegate;
```
rlepinski: do we need these fields?

crow: I'll remove them when I figure out why the strong wrapping isn't working. I wanted it to be in a working state.

PR #58, opened 2017-03-27
Strip debug symbols during iOS build

rlepinski:
Should address https://github.com/urbanairship/ios-library/issues/128


madelgi:
🌴

PR #62, opened 2017-04-07
MB-2098: Generate iOS bindings via diff

madelgi:
This didn't end up working well, but this generateBindings task may be useful in the future. I've also swapped out usages of `executable` in favor of `commandLine`


rlepinski:
👍

PR #74, opened 2017-10-02
Release 4.6.1

BrianBatchelder:
Fix deep link action and sample push handler issues
- ActionBlock had incorrect binding, causing a run-time exception.
- Sample Push Handler was not validating a pointer, causing an NPE.

Fix build.gradle to update version that was being missed.

Manually tested on iOS 11.

### DIFF
--- a/samples/ios-unified/PushHandler.cs
+++ b/samples/ios-unified/PushHandler.cs
@@ -83,9 +83,9 @@ namespace Sample
 			alertController.AddAction(okAction);
 
 			UIViewController topController = UIApplication.SharedApplication.KeyWindow.RootViewController;
-            if (alertController.PopoverPresentationController != null) {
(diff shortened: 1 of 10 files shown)


samples/ios-unified/PushHandler.cs
```diff
@@ -83,7 +83,9 @@ public override void ReceivedNotificationResponse(UANotificationResponse notific
 			alertController.AddAction(okAction);
 
 			UIViewController topController = UIApplication.SharedApplication.KeyWindow.RootViewController;
-			alertController.PopoverPresentationController.SourceView = topController.View;
+            if (alertController.PopoverPresentationController != null) {
```
rlepinski: Whitespace

marc-scig:
☮️

PR #75, opened 2017-10-09
Fix Xamarin Store submission issues.

BrianBatchelder:
There were two issues:

- "Firstly, with the release of iOS 11 the simulator, as well as real devices, will not launch 32bit only applications. In your iOS sample it is set to target only i386 instead of i386 – x86_64."
  - Fixed by adding x86_64 to the release build.
- "Secondly, the android sample has a missing dependency on CardView which means it won’t compile. Adding the nuget allows it to compile and run successfully."
  - Fixed by adding CardView to list of packages. A bunch of other packages updated automagically.

Note: all of the changes are to the sample code, so I will not be re-releasing the SDK. I will, however, advance the 4.6.1 tag in GitHub.


rlepinski:
👍

PR #78, opened 2017-10-25
Add UA notification service extension to sample app

BrianBatchelder:
Had to figure out to do this to answer a customer question about it. Might as well just put it in the repo. I also changed the bundle id to the same as our iOS samples, so I won't keep forgetting to change it when I'm testing.


rlepinski:
👍

marc-scig:
☯️

PR #80, opened 2017-12-05
Enable use of IUAInboxDelegate protocol

BrianBatchelder:
Changes the type of UAInbox delegates to the "IUAInboxDelegate" protocol, so classes that implement that protocol can be UAInboxDelegates. Doesn't require any source changes.


rlepinski:
👍

marc-scig:
🖼

PR #88, opened 2018-03-02
Update build tools

BrianBatchelder:
Updated command-line build tools to current Microsoft versions.
Addressed various build warnings.
- Missing & extraneous icons
- Mismatches in iOS app & service extension build configs
- Several issues with iOS bindings
- Gradle deprecation


samples/ios-unified/Sample.csproj
```diff
@@ -62,12 +62,13 @@
     <ConsolePause>false</ConsolePause>
     <MtouchArch>ARMv7, ARM64</MtouchArch>
     <CodesignEntitlements>Entitlements.plist</CodesignEntitlements>
-    <MtouchProfiling>true</MtouchProfiling>
+    <MtouchProfiling>false</MtouchProfiling>
```
marc-scig: What's this one about?

samples/ios-unified/SampleServiceExtension/SampleServiceExtension.csproj
```diff
@@ -70,7 +70,7 @@
     <MtouchDebug>true</MtouchDebug>
     <MtouchNoSymbolStrip>true</MtouchNoSymbolStrip>
     <MtouchFastDev>true</MtouchFastDev>
-    <MtouchFloat32>true</MtouchFloat32>
+    <MtouchFloat32>false</MtouchFloat32>
```
marc-scig: also curious about this one

src/AirshipBindings.iOS/ApiDefinition.cs
```diff
@@ -3107,7 +3108,7 @@ interface UAMessageCenterMessageViewController : IUAWKWebViewDelegate, UAMessage
     }
 
 	// @protocol UAMessageCenterMessageViewProtocol
-	[Protocol, Model]
+	[Protocol]
```
marc-scig: I've been following the `[Protocol, Model]` pattern in the tool I've been working on. What's the difference here, as opposed to all the other protocols we have bindings for?

src/AirshipBindings.iOS/ApiDefinition.cs
```diff
@@ -1396,6 +1396,7 @@ interface UADisposable
 
 		// -(void)dispose;
 		[Export ("dispose")]
+        [New]
```
marc-scig: What is this annotation for?

samples/ios-unified/Sample.csproj
BrianBatchelder: I can't remember exactly. It may have changed as a side-effect of another change I made.

samples/ios-unified/SampleServiceExtension/SampleServiceExtension.csproj
BrianBatchelder: This has to do with whether or not 32-bit floating point numbers are interpreted as 64-bit or something. I just matched what we had in our sample project.

src/AirshipBindings.iOS/ApiDefinition.cs
BrianBatchelder: I think it complained here because there was no base class. So I either had to add a base class or remove Model. The protocol doesn't really have a base class, so I chose to remove Model. Things seem to work.

src/AirshipBindings.iOS/ApiDefinition.cs
BrianBatchelder: So Xamarin adds a `Dispose()` to its version of NSObject. This says our `dispose()` is a new method, not an override.

marc-scig: ah interesting. I'm guessing this is just one of those things we need to special case, like "@object"

BrianBatchelder: Yes. Or maybe we can add a notation in the code that is picked up by Bluntie?

src/AirshipBindings.iOS/ApiDefinition.cs
marc-scig: Interesting. I just did some similar bluntie stuff with this very class, and its lack of a base class annotation. It's one of the few protocols in AirshipKit that doesn't explicitly implement the NSObject protocol. Which is not really a base class, but that's a whole other thing I guess.

src/AirshipBindings.iOS/ApiDefinition.cs
marc-scig: Perhaps. I think a reasonable pattern to follow is to special case things that we know ahead of time to conflict with c#, such as reserved keywords, and maybe introduce hints on the obj-c side for stuff that sourcekitten has difficulty parsing. But all suggestions are welcome.

BrianBatchelder: I think we are likely to get some stuff right and some stuff wrong at first. And some stuff will become right or wrong over time as our SDK and Xamarin change. But I'm so looking forward to being able to specify these things in some other way than manually editing `ApiDefinitions.cs` every time!!

marc-scig:
☯️

crow:
🍔 lgtm

PR #89, opened 2018-03-22
MOBILE-26 SDK 9.x

marc-scig:
This is a big one. I've tried to separate commits as best I can, so I recommend reviewing things more or less individually so as not to be overwhelmed or miss the forest for the trees. The largest single source of changes is the updated iOS bindings, which are worth some scrutiny, though so far I've had no difficulty testing with them. There are a lot more of them, because it now includes the entire (public) SDK. Keep an eye out for anything that looks weird and if in doubt, ask.

iOS sample code is not terribly different, though because we decided to stop translating accessor methods into properties there are number of superficial changes such as `UAirship.Push` -> `UAirship.Push()`, as well as interface changes at the SDK level between 8.x and 9.x. I added a blurb in the CHANGELOG about this.

The PCL was also rather out of date, having not been updated since component v 4.1. I changed that project to target our locally built nuget package so that it could point to the 5.0 component that hasn't been released yet. 

Lastly I'm least familiar with the android bindings, so this is worth a close look. The sample app is building and running fine now, so far as I can tell, though I had to do some shuffling of the Transforms/Metadata as well as LocationFragment.cs to reflect changes in our location API. I'm not entirely convinced it's the best approach, particularly where generics are concerned, but this might be the limit of my understanding at the moment, so suggestions or fixup commits are welcome.


src/AirshipBindings.Android/Additions/ILocationCallback.cs
```diff
@@ -1,139 +0,0 @@
-﻿/*
- Copyright 2017 Urban Airship and Contributors
-*/
-
-using System;
-using Android.Runtime;
-using Android.Locations;
-
-namespace UrbanAirship.Location
```
rlepinski: YES!

rlepinski:
Everything is looking pretty good. Make sure you can receive a push on Android. Take a look at https://github.com/urbanairship/xamarin-component/issues/77 for FCM issues when testing Android.

👍

PR #91, opened 2018-04-09
MOBILE-151 Update to latest SDKs, fix mtouch constants bug

marc-scig:
I wasn't able to reproduce the problem our customers have been having, but this should fix it. Re: https://urbanairship.atlassian.net/browse/CUSTENG-228


rlepinski:
👍

PR #92, opened 2018-04-18
MOBILE-153 .NET Standard Library

marc-scig:
I'm sure I'm forgetting something but might as well get the ball rolling.

For the uninitiated, .NET standard libraries are basically like Portable Class Libraries except that they target an actual standard instead of the morass of cobbled together platform extensions that PCL profiles have come to represent. This means little for our code in practice – we were never using platform specific stuff in the PCL to begin with – but it's Microsoft's chosen replacement for PCLs moving forward and I believe that apps that want to use .NET standard libs at all have to use them throughout, which is one of the reasons people have been requesting support for it.

.NET 2.0 is semi new, and support for creating apps consuming libraries targeting it it is not well cemented across Visual Studio versions yet. So I think we should stop short of officially deprecating the PCL just yet but I've done my best to call out the difference between the two in the docs and suggest devs use the new one if they have the option.

I made a basic forms app and tried a few operations to verify it was working as intended, such as displaying the message center. However my ability to test iOS is currently limited until I get on High Sierra and Xcode 9.3, so I could use some help with that if anyone is feeling adventurous.


rlepinski:
👍

docs/mainpage.md
```diff
@@ -1,19 +1,22 @@
-# AirshipBindings Portable Class Library
+# AirshipBindings .NET Standard Library and Portable Class Library
 
-The AirshipBindings portable client library exposes a common subset
-of functionality between the iOS and Android SDKs. The PCL can be used within
+The AirshipBindings .NET standard library and portable client library expose a common subset
```
BrianBatchelder: Suggestion: The AirshipBindings .NET standard ~library~ and portable client ~library~ libraries expose

PR #98, opened 2018-07-20
MOBILE-243 Update to iOS SDK 9.3.2, Android SDK 9.4 

marc-scig:
### What do these changes do?

Updates the Xamarin bindings to the latest SDK versions

Built the stuff, verified there were no errors or glaring issues. Ran the samples.

This is the second bindings release to use Xamboni, so take a closer look at the ApiDefinition.cs and StructsAndEnums.cs files to makes sure there's nothing crazy going on. It looks fine to my eye, though.

I also made a minor change in the sample app solution structure – the iOS SampleServiceExtension had an out of date dependency on the old app extensions package that caused it to fail to build. This likely slipped through when we were updating our versioning strategy. Instead of building against a package (local or remote) we should just include the app extensions project in the solution and build against a reference to its product, which is what the main iOS Sample is already doing.

### DIFF
--- a/samples/ios-unified/SampleServiceExtension/SampleServiceExtension.csproj
+++ b/samples/ios-unified/SampleServiceExtension/SampleServiceExtension.csproj
@@ -82,10 +82,6 @@
     <Reference Include="System.Xml" />
     <Reference Include="System.Core" />
     <Reference Include="Xamarin.iOS" />
-    <Reference Include="AirshipBindings.iOS.AppExtensions">
-      <HintPath>..\packages\urbanairship.ios.appextensions.5.1.0\lib\Xamarin.iOS10\AirshipBindings.iOS.AppExtensions.dll</HintPath>
-      <Private>False</Private>
-    </Reference>
   </ItemGroup>
   <ItemGroup>
     <Folder Include="Resources\" />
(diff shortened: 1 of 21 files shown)


samples/ios-unified/SampleServiceExtension/SampleServiceExtension.csproj
```diff
@@ -82,21 +82,22 @@
...
   </ItemGroup>
   <ItemGroup>
     <None Include="Info.plist" />
     <None Include="Entitlements.plist" />
-    <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="NotificationService.cs" />
   </ItemGroup>
+  <ItemGroup>
```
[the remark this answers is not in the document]
rlepinski: nvm, read the description

rlepinski:
👍

samples/ios-unified/SampleServiceExtension/SampleServiceExtension.csproj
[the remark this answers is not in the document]
BrianBatchelder: Nice catch!

src/AirshipBindings.Android.ADM/AirshipBindings.Android.ADM.csproj
```diff
@@ -90,7 +90,7 @@
     <Folder Include="Jars\" />
   </ItemGroup>
   <ItemGroup>
-    <LibraryProjectZip Include="Jars\urbanairship-adm-9.2.0.aar" />
+    <LibraryProjectZip Include="Jars\urbanairship-adm-9.4.0.aar" />
```
BrianBatchelder: I'm betting we can figure out a way to reference `airship.properties` or some other file so we don't have to hardcode the versions in the csproj files.

marc-scig: I think that's what happened here. I didn't change the reference, just updated airship.properties and ran the build script.

BrianBatchelder: I forgot about that. That's probably good enough.

PR #102, opened 2018-11-09
[MOBILE-376] update iOS and Android SDKs to 10.0.3, 9.5.4

marc-scig:
### What do these changes do?
Updates the iOS and Android SDKs to their latest patch.

There were some nasty bugs we patched recently, so this will spare our Xamarin users some difficulty.

I also updated the UrbanAirship.Core/FCM packages as well as the sample to build using the latest mono android (9.0) instead of pegging things to 8.1. I believe the mixture of 8.1 and 9.0 artifacts was causing some silent deployment errors resulting in Firebase woes, and we should probably be reflexively targeting the latest SDK version anyway.
  
Manually testing the sample app for iOS and Android.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ This library provides official bindings to the Urban Airship SDK, as well as sam
 
 ### Release Notes
 
+Version 9.3.1 – November 9, 2018
+================================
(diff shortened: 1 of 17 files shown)


rlepinski:
👍

PR #103, opened 2018-11-14
[android-9.5.5-patch] update android SDK to 9.5.5

marc-scig:
### What do these changes do?
Updates Android SDK to 9.5.5

9.5.5 fixes a fairly critical bug, so we want to make sure our framework users are covered.

Manual testing of the sample app. Nothing has changed beyond the version of the incorporated aar, and the android update itself is quite small.


rlepinski:
👍

PR #107, opened 2019-04-19
[release-9.4.1] release 9.4.1

marc-scig:
This is mostly a maintenance update to get our stuff happily building in Visual Studio 8/monoandroid 9. Package references have been updated to the latest available versions and weird crufty assembly errors resolved. I also went ahead and bumped iOS to 10.2.2 since carthage no longer works properly with earlier versions in Xcode 10.2.


rlepinski:
👍

PR #108, opened 2019-05-29
[MOBILE-635] Update iOS bindings to SDK 11.

BrianBatchelder:
### What do these changes do?
Update iOS bindings to iOS SDK 11.

Essential part of releasing the new version of the Xamarin Library.

Build and run.

- I renamed the ApiDefinition.cs files to the plural ApiDefinitions.cs. All the tools generate the second name. It is the standard.
- There will be more PRs before the 10.0 Xamarin library release.


samples/ios-unified/PushSettingsViewController.cs
```diff
@@ -20,7 +20,7 @@ public override void ViewDidLoad()
 
             // Initialize switches
             this.pushEnabledSwitch.On = UAirship.Push().UserPushNotificationsEnabled;
-            locationEnabledSwitch.On = UAirship.Location().LocationUpdatesEnabled;
+            locationEnabledSwitch.On = UAirship.Shared().LocationProviderDelegate.GetLocationUpdatesEnabled();
```
marc-scig: Did you get this working?

src/AirshipBindings.iOS/ApiDefinitions.cs
```diff
@@ -3302,20 +3757,94 @@ interface UAInAppMessageAdapterProtocol
 
     interface IUAInAppMessageAdapterProtocol { }
 
+    // @protocol UAInAppMessageCachePolicyDelegate <NSObject> /** * Return cache policy for caching assets on schedule * * @param message The message for which the assets will or won't be cached * @return `YES` requests the Asset Manager to cache the message's assets when the message is scheduled. * * @note If unimplemented, the message's assets will not be cached when the message is scheduled. */ - (BOOL)shouldCacheOnSchedule:(UAInAppMessage *)message; /** * Return cache policy for retaining cached assets after display * * @param message The message for which the assets will or won't be cached * @return `YES` requests the Asset Manager to persist the caching of the message's assets when * the message has finished displaying. * * @note If unimplemented, the message's assets will not be persisted when the message has finished displaying. */ - (BOOL)shouldPersistCacheAfterDisplay:(UAInAppMessage *)message; @end
```
marc-scig: looks like something mildly wonky is happening here with the generated comments in this section. I'd recommend hand editing any of the ones that run on weirdly like this, and make a note for sometime later to figure out what's happening. since xamboni gets that info more or less directly from sourcekitten, my best guess is that something may have changed in SourceKit/sourcekitten that causes this behavior, but it's hard to say without more info. not a high priority for now, beyond the formatting here

samples/ios-unified/PushSettingsViewController.cs
BrianBatchelder: Nah. It won't function until I add the LocationKit as a new package. That's my next PR.

PR #110, opened 2019-06-13
[MOBILE-636] Xamarin: Add new iOS Location module

BrianBatchelder:
### What do these changes do?
Add new iOS Location module/nuget

iOS SDK extracted the location code into a new LocationKit. This makes LocationKit available to Xamarin developers.

I added the new Location module to our Xamarin iOS Sample app. I tested that I could enable and disable location, that I saw the correct authorization dialogs, and the log showed location being received and a location event being created.


samples/ios-unified/AppDelegate.cs
```diff
@@ -14,6 +14,9 @@
 using System.Linq;
 using Xamarin.Forms;
 using System.Security.Authentication.ExtendedProtection;
+
+[assembly: Preserve (typeof (UALocation), AllMembers = true)]
```
marc-scig: if this is required to make the locationProviderDelegate work properly, would it be simpler to just use UALocation directly? Using the delegate in the native sample is really more of a hack, based on the awkwardness of dependencies that would have resulted with AirshipDebugKit.

src/AirshipBindings.iOS.LocationKit/ApiDefinitions.cs
```diff
@@ -0,0 +1,11 @@
+/*
+ Copyright Urban Airship and Contributors
+*/
+
+namespace UrbanAirship {
+
+    [BaseType(typeof(UAComponent))]
+    interface UALocation : IUALocationProviderDelegate {}
```
marc-scig: Are there no bindings for this?

BrianBatchelder: There are now.

samples/ios-unified/AppDelegate.cs
BrianBatchelder: Changed.

src/AirshipBindings.Portable/AirshipBindings.Portable.iOS/Airship.cs
```diff
@@ -52,25 +52,25 @@ public bool LocationEnabled
 		{
 			get
 			{
-				return UAirship.Location().LocationUpdatesEnabled;
+				return UAirship.Shared().LocationProviderDelegate.LocationUpdatesEnabled;
```
marc-scig: Does this actually work? I don't see any changes in here that make the portable library depend on the location module, so my guess is that the delegate will be null and these calls will crash. It's also worth asking if we even want the portable library to have such a dependency.

BrianBatchelder: Right. I have the PCL calling the delegate methods, so, if a user has their own location library, it will work. Sort of like we did in the native DebugKit. There will have to be docs for how to add location kit to PCL/.NET apps.

rlepinski: I want to avoid situations where everything will compile but silently fail at runtime. That will make it hard for people to figure out why things are broken. That is why we moved interface methods in cordova into its own package, and probably will do the same for react. 

Normally for the PCL, I would say make a dependency on the location module. However, since location is problematic for app store submission, we should just leave it out for now and if we get enough requests add back a PCL just for location calls.

marc-scig: I am kind of leaning away from that at this point. The location provider delegate is really mostly for the benefit of the SDK, not situations like these. It's probably my fault for using it in such a quick and dirty way in the react plugin, but the problem with this is that it increases the likelihood of someone getting an app that cleanly compiles but silently fails. Also if the integration story is cleaner for the full bindings but jankier for the PCL, that's kind of a red flag. I think our options here are either to remove it from the PCL for the time being or create a portable location submodule.

BrianBatchelder: Let me think about it a bit. Maybe we can talk about it during co-op time.

BrianBatchelder: Decision is to remove location from PCL and .NET. In progress.

PR #112, opened 2019-06-19
[MOBILE-700] Fix message center action.

BrianBatchelder:
### What do these changes do?
Fix the message center action (^mc) in the Xamarin Android sample app.

Keep stuff working.

Build and manual test.

https://urbanairship.slack.com/files/U3M8XNZ0T/FKS8X8QP8/message-center-action.webm

### DIFF
--- a/samples/android/UrbanAirshipListener.cs
+++ b/samples/android/UrbanAirshipListener.cs
@@ -4,15 +4,16 @@
 
 using System;
 using Android.Content;
-using Android.App;
+using Android.OS;
 using Android.Util;
-using UrbanAirship.Push;
+
 using UrbanAirship;
-using Android.Support.V4.Content;
+using UrbanAirship.Push;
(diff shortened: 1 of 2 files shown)


samples/android/UrbanAirshipListener.cs
```diff
@@ -68,6 +69,31 @@ public void OnNotificationDismissed(NotificationInfo notificationInfo)
 		{
 			Log.Info (TAG, "Notification dismissed. Alert: " + notificationInfo.Message.Alert + ". Notification ID: " + notificationInfo.NotificationId);
 		}
-	}
+
+        public bool OnShowMessageCenter(String messageId)
+        {
+            using (var h = new Handler(Looper.MainLooper))
```
rlepinski: Pretty sure this already runs on the main thread but its not annotated to do that. We should fix that in the next SDK release

PR #113, opened 2019-06-26
Xamarin: Fix push settings view

BrianBatchelder:
### What do these changes do?
Fix some problems with the display and control of push settings in the sample Settings view I found during testing.

Want to keep the sample apps working correctly.

Build & run on device.

This screenshot shows the authorizations displaying in the `Push Enabled` cell:
Screen Shot 2019-06-26 at 8 25 48 AM


BrianBatchelder:
What, you don't like to read 2-point type? ;-)
I think it is good enough for the sample app.

PR #121, opened 2019-12-09
MOBILE-1119 Update Xamarin iOS to bind to SDK 12.1.2

marc-scig:
This update Xamarin to use iOS SDK 12.1.2, and bumps the iOS nuspecs to 12.1.0. Everything else remains as is.

The changes in the bindings diff are mostly harmless, and much of it is just whitespace/line noise as a result of minor changes to the sourcekitten output for method prototypes. The meaningful stuff is in the attribute mutations.

I tested by running the sample app and verifying that nothing is wonky.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "urbanairship/ios-library" == 12.0.0
+github "urbanairship/ios-library" == 12.1.2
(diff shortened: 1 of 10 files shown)


rlepinski:
👍

PR #122, opened 2020-01-27
MOBILE-1046: Update iOS to SDK 13

BrianBatchelder:
### What do these changes do?
This PR updates the iOS bindings to support SDK 13.

Version 13 modularized the SDK, which requires corresponding changes to the Xamarin iOS bindings.

Tested channel registration, push, message center and notification service extension on iOS 13 device. Tested build scripts.

- The work also required some minor changes to the iOS SDK, mostly added doc comments for elements that are internal but weren't marked that way. These changes should not require an iOS re-release. They will follow in a separate PR.  
- The work also required enhancements to Xamboni. That will also follow in a separate PR.


UrbanAirship.iOS.Core.nuspec
```diff
@@ -1,16 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
```
rlepinski: I think we still want a UrbanAirship.iOS.nuspec that depends on Core, Automation, MessageCenter, and ExtendedActions

src/AirshipBindings.iOS.Automation/build.gradle
```diff
@@ -28,7 +28,7 @@ task generateBindings {
```
rlepinski: Is this task still valid? we generate bindings with Xamboni now

BrianBatchelder: It is not valid. I could remove.

UrbanAirship.iOS.Core.nuspec
BrianBatchelder: OK.

src/AirshipBindings.Android.ADM/build.gradle
```diff
@@ -27,7 +27,7 @@ task syncVersion  {
         def path = new File("$projectDir/Jars/$filename")
         if ( !path.exists() ) {
            // Download aar
-            def url = new URL("http://central.maven.org/maven2/com/urbanairship/android/urbanairship-adm/${airshipProperties.androidVersion}/$filename")
+            def url = new URL("https://repo1.maven.org/maven2/com/urbanairship/android/urbanairship-adm/${airshipProperties.androidVersion}/$filename")
```
rlepinski: you should use the jcenter repos

src/AirshipBindings.iOS.Core/ApiDefinitions.cs
```diff
@@ -0,0 +1,2852 @@
...
+        [Export("consumedTemplateWithValue:")]
+        UAMediaEventTemplate ConsumedTemplate ([NullAllowed] NSNumber eventValue);
+
+        // - (nonnull UACustomEvent *)createEvent;
+        [Export("createEvent")]
+        UACustomEvent CreateEvent ();
+    }
+
+    // @interface UAModifyTagsAction : UAAction
+    [BaseType(typeof(UAAc
```
marc-scig: Can you do a search/replace for "Deprecated ?" -> "Deprecated -"

There's a ticket to fix the issue that's currently making the question mark appear.

src/AirshipBindings.Android.ADM/build.gradle
marc-scig: What was the reason for this change? Ryan thinks we should probably use jcenter over maven, though that's probably something we can do later.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/AirshipBindings.NETStandard.Android.csproj
```diff
@@ -14,7 +14,6 @@
     <AndroidResgenClass>Resource</AndroidResgenClass>
     <MonoAndroidResourcePrefix>Resources</MonoAndroidResourcePrefix>
     <MonoAndroidAssetsPrefix>Assets</MonoAndroidAssetsPrefix>
-    <AndroidUseLatestPlatformSdk>true</AndroidUseLatestPlatformSdk>
```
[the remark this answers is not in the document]
BrianBatchelder: It looks like it has been [deprecated](https://forums.xamarin.com/discussion/158550/androiduselatestplatformsdk-not-recognised-by-vs-mac-2019). Must have been an automatic Visual Studio change. Same change was made to `AirshipBindings.Portable.Android.csproj`

src/AirshipBindings.iOS.Core/ApiDefinitions.cs
BrianBatchelder: Good catch. Forgot about that one.

src/AirshipBindings.Android.ADM/build.gradle
BrianBatchelder: Maven Central no longer allows `http:`. This is the recommended URL. But Ryan wanted bintray, so I've changed it to bintray.

BrianBatchelder: Maven Central no longer allows `http:`. This is the recommended URL. But Ryan wanted bintray, so I've changed it to bintray.

src/AirshipBindings.iOS.Core/ApiDefinitions.cs
BrianBatchelder: FYI - I checked the iOS SDK source, and it does indeed contain a unicode "-" which Xamboni converts into that "?". We should probably have a ticket to clean unicode out of the SDK.

marc-scig: There's already a ticket for that.

PR #123, opened 2020-02-04
[MOBILE-1294] Update Xamarin to iOS 13.1

BrianBatchelder:
### What do these changes do?
Update Xamarin iOS bindings to 13.1

Keep Xamarin up-to-date.

Local builds and tests.

Due to the modularization and maybe data collection features, I think we need to update other documentation for this release, e.g.
- migration guide
- extdocs


src/AirshipBindings.iOS.MessageCenter/ApiDefinitions.cs
```diff
@@ -240,6 +240,7 @@ interface UAMessageCenter
         // + (null_unspecified instancetype)shared;
         [Static]
         [Export("shared")]
+        [New]
```
marc-scig: what's the deal with these `[New]` annotations?

BrianBatchelder: `[New]` clears up a warning in the build. Basically, the [warning](https://stackoverflow.com/questions/19193821/use-the-new-keyword-if-hiding-was-intended-warning) is "Shared() hides inherited member Shared(). Use the new keyword if hiding was intended.". The problem is that we need the subclass `Shared()` methods, as they have the correct return type. If we don't have them, then the calling C# has to cast UAComponent to the correct return type.

marc-scig: got it, thanks

UrbanAirship.NETStandard.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
    <metadata>
       <id>urbanairship.netstandard</id>
-      <version>11.1.0</version>
+      <version>12.0.0</version>
```
marc-scig: what are the rules we're following for updating the versions for netstandard and portable again? I'm confused why this is a major version bump.

BrianBatchelder: For .NETStandard and PCL, we bump it at least as much as we bumped the platform-specific. So, if iOS got a major and Android got a minor, .NETStandard and PCL get a major.

marc-scig: But it was at 11.1.0 for 13.0.4?

marc-scig: was just just an oversight last time around?

BrianBatchelder: Sort of an oversight. I didn't go through all of the release steps last time around because we didn't release.

PR #125, opened 2020-02-07
MOBILE-1303

marc-scig:
There were a few constants that were causing linker errors when building against the nupkg distribution. This doesn't seem to apply when testing using the "library project" sample, so we need to be more careful about testing using real packages in the future.

Nearly all of these are constants whose implementations have been removed from the SDK but for whatever reason didn't get removed from the headers. Two of them, namely UANotificationOptionNone and UAInAppMessageDisplayCoordinatorIsReadyKey are still in the SDK, but have inline definitions in the header file. We should probably also break the habit of doing this in the SDK, and Xamarin to my knowledge doesn't support binding them. I removed the latter for now, which is almost certainly something nobody will miss.

While I was testing I also noticed that several of the iOS nupkgs had out of date dependencies. VS was warning about this locally, because it matches the depdendency as >= and was selecting the latest version on disk. A remote package install probably would have resulted in silently installing the wrong dependency. It turned out build.gradle was doing the regex incorrectly, so I fixed the script as well.

### DIFF
--- a/UrbanAirship.iOS.nuspec
+++ b/UrbanAirship.iOS.nuspec
@@ -11,10 +11,10 @@
 
       <dependencies>
          <group targetFramework="Xamarin.iOS">
-            <dependency id="urbanairship.ios.core" version="13.0.4"/>
-            <dependency id="urbanairship.ios.automation" version="13.0.4"/>
-            <dependency id="urbanairship.ios.extendedactions" version="13.0.4"/>
-            <dependency id="urbanairship.ios.messagecenter" version="13.0.4"/>
+            <dependency id="urbanairship.ios.core" version="13.1.0"/>
+            <dependency id="urbanairship.ios.automation" version="13.1.0"/>
+            <dependency id="urbanairship.ios.extendedactions" version="13.1.0"/>
+            <dependency id="urbanairship.ios.messagecenter" version="13.1.0"/>
          </group>
       </dependencies>
    </metadata>
--- a/build.gradle
+++ b/build.gradle
@@ -137,8 +137,8 @@ task syncVersion {
 
         // handle urbanairship.ios dependencies
         ant.replaceregexp(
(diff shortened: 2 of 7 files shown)


UrbanAirship.iOS.nuspec
```diff
@@ -11,10 +11,10 @@
...
       <dependencies>
          <group targetFramework="Xamarin.iOS">
-            <dependency id="urbanairship.ios.core" version="13.0.4"/>
-            <dependency id="urbanairship.ios.automation" version="13.0.4"/>
-            <dependency id="urbanairship.ios.extendedactions" version="13.0.4"/>
-            <dependency id="urbanairship.ios.messagecenter" version="13.0.4"/>
+            <dependency id="urbanairship.ios.core" version="13.1.0"/>
+            <dependency id="urbanairship.ios.automation" version="13.1.0"/>
+            <dependency id="urbanairship.ios.extendedactions" version="13.1.0"/>
+            <dependency id="urbanairship.ios.messagecenter" version="13.1.0"/>
```
BrianBatchelder: nice catch!

BrianBatchelder: nice catch!

build.gradle
```diff
@@ -137,8 +137,8 @@ task syncVersion {
 
         // handle urbanairship.ios dependencies
         ant.replaceregexp(
-            match: "<dependency id=\"urbanairship.ios\" version=\".*\"/>",
-            replace: "<dependency id=\"urbanairship.ios\" version=\"$iosNugetVersion\"/>",
+            match: "<dependency id=\"urbanairship.ios(.*)\" version=\".*\"/>",
+            replace: "<dependency id=\"urbanairship.ios\\1\" version=\"$iosNugetVersion\"/>",
```
BrianBatchelder: And future thanks!

PR #130, opened 2020-03-12
[MOBILE-1222] .NETStandard/portable attribute support

marc-scig:
This adds attribute support to the .NETStandard and Portable libraries. The pattern is similar to the way we ended up implementing CustomEvent properties recently, with generics and exhaustive type checking instead of Object casts, which are illegal in C# in situations that would seem reasonable for Java. Luckily C# has enough syntactic sugar to make it reasonably painless.

On iOS and Android I was able to set attributes from an ad-hoc test app, using locally built nuget packages, and verify they were being bridged properly to the SDK.

Note: VS will complain about missing references and ask you to install a variety of androidx packages by hand unless you change the target SDK version of the forms app to API 29.

### DIFF
--- a/UrbanAirship.Android.Preference.nuspec
+++ b/UrbanAirship.Android.Preference.nuspec
@@ -11,6 +11,7 @@
 
       <dependencies>
          <group targetFramework="MonoAndroid">
+            <dependency id="Xamarin.AndroidX.Preference" version="1.0.0"/>
             <dependency id="urbanairship.android.core" version="12.2.0"/>
          </group>
       </dependencies>
--- a/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
+++ b/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
@@ -3,6 +3,7 @@
 */
 
 using System.Collections.Generic;
+using UrbanAirship.NETStandard.Attributes;
 
 namespace UrbanAirship.NETStandard
 {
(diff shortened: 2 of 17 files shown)


src/AirshipBindings.NETStandard/AirshipBindings.NETStandard/Airship.cs
```diff
@@ -123,6 +123,16 @@ public int MessageCenterCount
...
         }
 
+        /// <summary>
+        /// Edit channel attributes.
+        /// </summary>
+        /// <returns>An <see cref="UrbanAirship.Portable.Attributes.AttributeEditor">AttributeEditor</see>
+        /// for channel attributes.</returns>
+        public Attributes.AttributeEditor EditAttributes()
+        {
+            throw new NotImplementedException(BaitWithoutSwitchMessage);
```
rlepinski: lol `BaitWithoutSwitch`

crow:
how do we test this/know all this builds properly without having a cross platform sample?

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard/Airship.cs
marc-scig: I know, I know

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -172,6 +173,49 @@ public Channel.TagGroupsEditor EditChannelTagGroups()
             });
         }
 
+        public AttributeEditor EditAttributes()
+        {
+            return new AttributeEditor((List<AttributeEditor.IAttributeOperation> operations) =>
+            {
+                var editor = UAirship.Shared().Channel.EditAttributes();
```
crow: UAChannel.Shared().something?

marc-scig: this one is for android, so that looks right to me

UrbanAirship.Android.Preference.nuspec
```diff
@@ -11,6 +11,7 @@
 
       <dependencies>
          <group targetFramework="MonoAndroid">
+            <dependency id="Xamarin.AndroidX.Preference" version="1.0.0"/>
```
BrianBatchelder: I think version 1.1.0 is available.

marc-scig: it is, but this is the version urbanairship-preference currently targets.

PR #131, opened 2020-03-13
[MOBILE-1386] CI

rlepinski:
CI to github actions

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,8 +19,10 @@ jobs:
           brew install graphviz
 
       - name: Create google-services.json
+        env:
+          GOOGLE_SERVICES_JSON: ${{ secrets.GOOGLE_SERVICES_JSON }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,28 @@
...
+          brew install doxygen
+          brew install graphviz
+
+      - name: Create google-services.json
+        env:
+          GOOGLE_SERVICES_JSON: ${{ secrets.GOOGLE_SERVICES_JSON }}
+        run: |
+          echo "$GOOGLE_SERVICES_JSON" > samples/android/Assets/google-services.json
+
+      - name: Build
```
crow: we don't want to run the samples?

BrianBatchelder: `./gradlew build` does build the samples. It doesn't create the NuGet Packages, I think.

rlepinski: build should cover all. the current ci is only building samples so this is doing more. Ill look into the nuget package parts

PR #132, opened 2020-03-13
[MOBILE-1387] Move release to github actions

rlepinski:
Cleaned up changelog as well. Going forward we describe our releases in terms of the PCL/NetStandard and only tag those, bumping binding versions is a byproduct of that. To trigger a release, push the version of the netstandard library. This next release I assuming it will be `13.0.0`


build.gradle
```diff
@@ -240,3 +240,11 @@ build.dependsOn(':docs:build',
     ':src:AirshipBindings.iOS.Location:build',
     ':src:AirshipBindings.iOS.NotificationContentExtension:build',
     ':src:AirshipBindings.iOS.NotificationServiceExtension:build' )
+
+
+
```
BrianBatchelder:
```suggestion
```

PR #137, opened 2020-04-01
[MOBILE-1142] Add plugin tracking for iOS

BrianBatchelder:
### What do these changes do?
Add plugin tracking for iOS to Xamarin component. Now Xamarin iOS apps will report something like `"X-UA-Frameworks" = "xamarin:13.1.1.0";` to analytics when they run.

Better understanding of the usage of our Xamarin components.

Ran on device with trace-level logging to see headers uploaded to Analytics.

The `registerSDKExtension:version:` method and `UASDKExtension` enum are not exposed through our API bindings. They are labelled `:nodoc:` in the iOS SDK and Xamboni ignores `:nodoc:` items, assuming they are SDK-private. I temporarily exposed the two and ran Xamboni to generate the appropriate bindings, which I then copied into `UAirship.cs`, my class that actually registers the sdk extension. It's kinda ugly, but works just fine. If we choose to expose the method and enum in the future, the `AirshipBindings.iOS.Core` build will fail, but all we have to do is remove the bindings I copied into `UAirship.cs`.

### DIFF
--- a/src/AirshipBindings.iOS.Core/Additions/UAirship.cs
+++ b/src/AirshipBindings.iOS.Core/Additions/UAirship.cs
@@ -1,0 +1,72 @@
+﻿/*
+ Copyright Airship and Contributors
+*/
+
+using System;
+using System.Reflection;
+using Foundation;
+using ObjCRuntime;
+
+namespace UrbanAirship
+{
+    public partial class UAirship
+    {
+        static UAirship()
+        {
+            NSNotificationCenter.DefaultCenter.AddObserver(new NSString("com.urbanairship.airship_ready"), (notification) =>
+            {
+                // Register Airship Xamarin component
+                Version version = Assembly.GetExecutingAssembly().GetName().Version;
+                UAirship.Analytics().RegisterSDKExtension(UASDKExtension.Xamarin, version.ToString());
+            });
+
+        }
+    }
+
+    // SDK-private enum and method declarations
+    [Native]
+    internal enum UASDKExtension : ulong
+    {
+        Cordova = 0,
+        Xamarin = 1,
+        Unity = 2,
+        Flutter = 3,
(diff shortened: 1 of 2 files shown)


src/AirshipBindings.iOS.Core/Additions/UAirship.cs
```diff
@@ -0,0 +1,72 @@
...
+
+    // SDK-private enum and method declarations
+    [Native]
+    internal enum UASDKExtension : ulong
+    {
+        Cordova = 0,
+        Xamarin = 1,
+        Unity = 2,
+        Flutter = 3,
+        ReactNative = 4
```
BrianBatchelder, line 34: Since Xamarin assigns fixed integer values to each of the enum values, I think we should do the same in the SDK to avoid any accidental changes through re-ordering or additions. Thoughts?

rlepinski: sure, create a ticket

src/AirshipBindings.iOS.Core/Additions/UAirship.cs
```diff
@@ -0,0 +1,72 @@
...
+using System;
+using System.Reflection;
+using Foundation;
+using ObjCRuntime;
+
+namespace UrbanAirship
+{
+    public partial class UAirship
+    {
+        static UAirship()
```
rlepinski, line 14: You should be able to do this exact thing on Android. Use the shared with the onready callback

src/AirshipBindings.iOS.Core/Additions/UAirship.cs
BrianBatchelder: https://urbanairship.atlassian.net/browse/MOBILE-1429

PR #138, opened 2020-04-02
[MOBILE-1392] add plugin tracking for android

crow:
### What do these changes do?
Adds plugin tracking to android

Was able to get it to break on the correct call on sample launch:
image

### DIFF
--- a/src/AirshipBindings.Android.Core/Additions/UAirship.cs
+++ b/src/AirshipBindings.Android.Core/Additions/UAirship.cs
@@ -6,6 +6,8 @@ using System;
 
 using Android.App;
 using Android.OS;
+using System.Reflection;
+using UrbanAirship.Analytics;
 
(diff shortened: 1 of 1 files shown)


src/AirshipBindings.Android.Core/Additions/UAirship.cs
```diff
@@ -47,6 +47,10 @@ public AirshipReadyCallback(Action<UAirship> callback)
 			}
 
 			public void OnAirshipReady (UAirship airship) {
+				Version version = Assembly.GetExecutingAssembly().GetName().Version;
```
rlepinski, line 50: This only works if they call takeOff or shared on UAirship directly. I think you want to have:

static UAirship {
  UAirship.shared( airship -> {
     airship.Analytics.RegisterSDKExttension(...);
   });
}

PR #139, opened 2020-04-03
[MOBILE-1142] Update plugin tracking to track the cross-platform version rather than the platform-dependent version

BrianBatchelder:
### What do these changes do?
Change the plugin tracking to report the cross-platform version rather than the platform-dependent version.

We want the common version

Local testing.

- Added a new `UACrossPlatformVersion` attribute in a new common `assemblyinfo.cs`.
- Extended build.gradle to update the new `UACrossPlatformVersion` attribute.
- Changed the Android and iOS code to use the new cross platform version rather than the platform specific version.
- Cleaned up all the projects to remove empty `AssemblyInfo.cs` files and make sure all sub-projects include the appropriate `SharedAssemblyInfo.cs`.
- I also set the `airship.properties` `crossPlatformVersion` back to 13.0.0 so it matched up with the versions elsewhere in the repo.


airship.properties
```diff
@@ -1,3 +1,3 @@
-crossPlatformVersion = 12.0.1
+crossPlatformVersion = 13.0.0
```
crow, line 1: did this need to change? it's probably going to conflict with the release branch

src/AirshipBindings.Android.Core/Additions/UAirship.cs
```diff
@@ -24,8 +25,15 @@ static UAirship()
...
-				airship.Analytics.RegisterSDKExtension(AnalyticsClass.ExtensionXamarin, version.ToString());
+				// Register Airship Xamarin component
+				Object[] crossPlatformVersions = Assembly.GetExecutingAssembly().GetCustomAttributes(typeof(UACrossPlatformVersionAttribute), false);
+				if (crossPlatformVersions.Length >= 1)
+				{
+					UACrossPlatformVersionAttribute version = Assembly.GetExecutingAssembly().GetCustomAttributes(typeof(UACrossPlatformVersionAttribute), false)[0] as UACrossPlatformVersionAttribute;
+					airship.Analytics.RegisterSDKExtension(AnalyticsClass.ExtensionXamarin, version.Version.ToString());
+				}
+
+				//airship.Analytics.RegisterSDKExtension(AnalyticsClass.ExtensionXamarin, version.ToString());
```
crow, line 36: commented out code

BrianBatchelder: Oops.

airship.properties
marc-scig: No need to change this here

PR #142, opened 2020-05-04
[MOBILE-24] Add deep link event listener

oristanovic:
### What do these changes do?
Add a deep link event listener to the NetStandard lib, mirrors the event listener in Cordova/React Native.

Created a xamarin forms app and tested manually on iOS and Android, checked via logs that the event callback is executed with the correct parameters.


src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/IAirship.cs
```diff
@@ -53,5 +53,7 @@ int MessageCenterCount
         Channel.TagGroupsEditor EditChannelTagGroups();
         
         Attributes.AttributeEditor EditAttributes();
+
+        void RegisterEventListener(Action<string, Dictionary<string, string>> onEventReceived);
```
rlepinski, line 57: C# has an event listener pattern that we should follow:

```
        public delegate void DeepLinkHandler (string deepLink);

        public event DeepLinkHandler OnDeepLink;
``` 

Then you add/remove using:
```
UAirship.instance.OnDeepLink += OnDeepLink;
UAirship.instance.OnDeepLink -= OnDeepLink;
```

oristanovic: Noice, updated the listener.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/Analytics/DeepLinkEvent.cs
```diff
@@ -0,0 +1,39 @@
+﻿/*
+ Copyright Airship and Contributors
+*/
+
+using System.Collections.Generic;
+
+namespace UrbanAirship.NETStandard.Analytics
```
rlepinski, line 7: do we need this still?

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -63,6 +66,22 @@ public string NamedUser
...
+            add
+            {
+                onDeepLinkReceived += value;
+                UAirship.Shared().WeakDeepLinkDelegate = this;
+            }
+
+            remove
+            {
+                onDeepLinkReceived -= value;
+                UAirship.Shared().WeakDeepLinkDelegate = null;
```
rlepinski, line 81: do we need to see if the total listeners is empty before removing the delegate?

PR #144, opened 2020-05-05
[MOBILE-1486] MessagePage

marc-scig:
This PR adds a Forms control deriving from `ContentPage` that allows easy loading of message center messages in a cross platform app. ContentPages in Forms are analogous to view controllers, and literally descend from `UIViewController` on iOS.

One of the consequences of this is that our NETStandard library will now have an explicit dependency on Xamarin.Forms for the first time. We also don't currently have NETStandard bindings for anything message center related, such as fetching and access to the message list, so that will need to be done in another ticket if this is going to be useful to anyone.

Here's an example of how to use it

```               
var messagePage = new MessagePage();

// Load the message
messagePage.MessageId = "SOME_MESSAGE_ID";

// Sign up for events
messagePage.Loaded += (object obj, MessageLoadedEventArgs ea) =>
{
    // Do something
};

messagePage.LoadFailed += (object obj, MessageLoadFailedEventArgs ea) =>
{
    // Do something
};

messagePage.Closed += (object obj, MessageClosedEventArgs ea) =>
{
    // Do something
};

// Push it onto a navigation stack (or whatever)
await Navigation.PushAsync(messagePage);
```

Doing this cross platform means creating a custom `PageRenderer` for both iOS an Android, which handles the platform-specific stuff without the MessagePage class having to know anything about it. The renderer is bound to the page using assembly annotations, and there are subtle challenges for iOS when doing that in a DLL, which I'll explain more about in the diff.

Tested in ad ad-hoc forms app on both platforms, with messages containing action buttons, to verify the native bridge is working correctly.

Simulator Screen Shot - iPhone 11 - 2020-05-05 at 16 04 59

Screenshot_1588719883

### DIFF
--- a/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/Message Center/MessagePage.cs
+++ b/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/Message Center/MessagePage.cs
@@ -1,0 +1,153 @@
+﻿/*
+ Copyright Airship and Contributors
+*/
+
+using System;
+
+using Xamarin.Forms;
+
+namespace UrbanAirship.NETStandard.MessageCenter
+{
+    /// <summary>
+    /// ContentPage used for displaying Message Center Messages.
+    /// </summary>
+    public partial class MessagePage : ContentPage
+    {
+        /// <summary>
+        /// Gets and sets the message ID.
+        /// </summary>
+        /// <value>The message ID.</value>
+        public string MessageId { get; set; }
+
+        /// <summary>
+        /// Event handler invoked when the message has loaded.
+        /// </summary>
+        public event EventHandler<MessageLoadedEventArgs> Loaded;
+
+        /// <summary>
+        /// Event handler invoked whe  the message load fails.
+        /// </summary>
+        public event EventHandler<MessageLoadFailedEventArgs> LoadFailed;
+
+        /// <summary>
+        /// Event handler invoked when the message is closed.
+        /// </summary>
+        public event EventHandler<MessageClosedEventArgs> Closed;
+
+        /// <summary>
+        /// MessagePage constructor.
+        /// </summary>
+        public MessagePage()
+        {
+        }
+
+        //@cond IGNORE
+        public void OnRendererLoaded(string messageId)
+        {
+            Loaded?.Invoke(this, new MessageLoadedEventArgs(messageId));
+        }
+
+        public void OnRendererLoadFailed(string messageId, bool retriable, MessageFailureStatus status)
+        {
+            LoadFailed?.Invoke(this, new MessageLoadFailedEventArgs(messageId, retriable, status));
+        }
+
+        public void OnRendererClosed(string messageId)
+        {
+            Closed?.Invoke(this, new MessageClosedEventArgs(messageId));
+        }
+        //@endcond
+    }
+
+    /// <summary>
+    /// Event handler invoked when the message is closed.
+    /// </summary>
+    public enum MessageFailureStatus
+    {
+        /// <summary>
+        /// The message is unavailable.
+        /// </summary>
+        Unavailable,
+
+        /// <summary>
+        /// The message list could not be fetched.
+        /// </summary>
+        FetchFailed,
+
+        /// <summary>
+        /// The message failed to load.
+        /// </summary>
+        LoadFailed
+    }
+
+    /// <summary>
+    /// Arguments for the LoadFailed event.
+    /// </summary>
+    public class MessageLoadFailedEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Gets the retriable value.
+        /// </summary>
+        /// <value>Whether the message load is retriable.</value>
+        public bool Retriable { get; private set; }
+
+        /// <summary>
+        /// Gets the message failure status.
+        /// </summary>
+        /// <value>The message failure status</value>
+        public MessageFailureStatus FailureStatus { get; private set; }
+
+        /// <summary>
+        /// Gets the message ID.
+        /// </summary>
+        /// <value>The message ID.</value>
+        public string MessageId { get; private set; }
+
+        //@cond IGNORE
+        public MessageLoadFailedEventArgs(string messageId, bool retriable, MessageFailureStatus status)
+        {
+            MessageId = messageId;
+            Retriable = retriable;
+            FailureStatus = status;
+        }
+        //@endcond
+    }
+
+    /// <summary>
+    /// Arguments for the Loaded event.
+    /// </summary>
+    public class MessageLoadedEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Gets the message ID.
+        /// </summary>
+        /// <value>The message ID.</value>
+        public string MessageId { get; private set; }
+
+        //@cond IGNORE
+        public MessageLoadedEventArgs(string messageId)
+        {
+            MessageId = messageId;
+        }
+        //@endcond
+    }
+
+    /// <summary>
+    /// Arguments for the Closed event.
+    /// </summary>
+    public class MessageClosedEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Gets the message ID.
+        /// </summary>
+        /// <value>The message ID.</value>
+        public string MessageId { get; private set; }
+
+        //@cond IGNORE
+        public MessageClosedEventArgs(string messageId)
+        {
+            MessageId = messageId;
+        }
+        //@endcond
+    }
+}
--- a/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/Message Center/MessagePage.cs
+++ b/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/Message Center/MessagePage.cs
@@ -13,11 +13,29 @@ namespace UrbanAirship.NETStandard.MessageCenter
     /// </summary>
     public partial class MessagePage : ContentPage
     {
+        private string messageId;
+
+        public static readonly BindableProperty MessageIdProperty = BindableProperty.Create(
+            propertyName: "MessageId",
+            returnType: typeof(string),
+            declaringType: typeof(MessagePage),
+            defaultValue: null,
+            defaultBindingMode: BindingMode.OneWay,
+            propertyChanged: null);
+
         /// <summary>
         /// Gets and sets the message ID.
         /// </summary>
         /// <value>The message ID.</value>
-        public string MessageId { get; set; }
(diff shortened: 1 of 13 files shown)


src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/Message Center/MessagePage.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        /// <summary>
+        /// MessagePage constructor.
+        /// </summary>
+        public MessagePage()
+        {
+        }
+
+        //@cond IGNORE
+        public void OnRendererLoaded(string messageId)
```
marc-scig, line 45: These methods were an attempt to bridge the more c-sharp-y EventHandlers in the MessagePage with the callbacks in the renderers. C# is very strict about the invocation of event handlers and doesn't allow them to be invoked outside of the scope they're defined in. Since renders can access the page at runtime but not the other way around, this seems like a reasonable compromise.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/MessagePageRenderer.cs
```diff
@@ -0,0 +1,166 @@
...
+using UrbanAirship.Widget;
+using Xamarin.Forms.Platform.Android;
+
+namespace UrbanAirship.NETStandard.Android
+{
+    public class MessagePageRenderer : PageRenderer, RichPushInbox.IFetchMessagesCallback
+    {
+        private class MessageWebViewClient : UAWebViewClient
+        {
+            private MessagePageRenderer renderer;
```
marc-scig, line 17: This might seem circular, but it's the only way to get access to variables in the outer scope, since C# doesn't have an equivalent of anonymous inner classes as in Java

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/MessagePageRenderer.cs
```diff
@@ -0,0 +1,166 @@
...
+                {
+                    renderer.messagePage.OnRendererLoadFailed(renderer.message.MessageId, false, MessageFailureStatus.LoadFailed);
+                    return;
+                }
+
+                renderer.message.MarkRead();
+                renderer.messagePage.OnRendererLoaded(renderer.message.MessageId);
+            }
+
+#pragma warning disable CS0672 // Member overrides obsolete member
```
marc-scig, line 44: these warnings are about deprecated methods we still have to use if we're targeting < API 23

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/MessagePageRenderer.cs
```diff
@@ -0,0 +1,166 @@
...
+                return;
+            }
+
+            messagePage = e.NewElement as MessagePage;
+            messageId = messagePage.MessageId;
+
+            LoadMessage(messageId);
+        }
+
+        protected override void OnLayout(bool changed, int l, int t, int r, int b)
```
marc-scig, line 136: This was a non-obvious hack discovered in the forums to get around an apparent Xamarin.Forms bug giving the view the wrong dimensions at runtime

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/AirshipForms.cs
```diff
@@ -0,0 +1,18 @@
...
+ Copyright Airship and Contributors
+*/
+
+using System;
+namespace UrbanAirship.NETStandard
+{
+    /// <summary>
+    /// Entry point for initializing Airship Forms controls.
+    /// </summary>
+    public class AirshipForms
```
marc-scig, line 11: This is only needed on iOS. In order for the right binding to be made between the page and renderer, the assembly containing the renderer must be loaded at the time Xamarin.Forms is initialized in the platform specific version of the App. On iOS this isn't the case by default, so this class is just here with a dummy Init method to force that to happen. You have to do this in your app delegate before anything else, like below. Otherwise nothing will happen when you load the page.

```
public override bool FinishedLaunching(UIApplication app, NSDictionary options)
{
    UrbanAirship.NETStandard.AirshipForms.Init();
    global::Xamarin.Forms.Forms.Init();
     // ...
}
```
This isn't documented behavior, but rather community knowledge shared in the forums. We'll need to make this clear in the docs.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/MessagePageRenderer.cs
```diff
@@ -0,0 +1,166 @@
...
+        private string messageId;
+        private FrameLayout nativeView;
+
+        public MessagePageRenderer(Context context) : base(context)
+        {
+            webView = new UAWebView(Context);
+            webViewClient = new MessageWebViewClient(this);
+
+            var fl = new FrameLayout(Context);
+            fl.LayoutParameters = new LayoutParams(LayoutParams.MatchParent, LayoutParams.MatchParent);
```
rlepinski, line 82: Might not actually be needed to wrap in a framelayout. I had to do that it react because of a react bug. Doesnt hurt though.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/MessagePageRenderer.cs
```diff
@@ -0,0 +1,145 @@
...
+                });
+            }
+        }
+
+        public void Close()
+        {
+            messagePage.OnRendererClosed(messageId);
+        }
+
+        [Export("webView:decidePolicyForNavigationResponse:decisionHandler:")]
```
marc-scig, line 72: These export annotations were added by VS automatically at its own suggestion, and my assumption is that they're needed because the bound delegate protocol diverges significantly from the original selector and messes up introspection on the Obj-c side. But that's just a guess.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/Message Center/MessagePage.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// <summary>
+        /// Gets and sets the message ID.
+        /// </summary>
+        /// <value>The message ID.</value>
+        public string MessageId { get; set; }
+
+        /// <summary>
+        /// Event handler invoked when the message has loaded.
+        /// </summary>
+        public event EventHandler<MessageLoadedEventArgs> Loaded;
```
rlepinski, line 25: You might want to expose LoadStarted to make using the MessagePage easier, technically they should know when we trigger a reload, but this way they can sign up a loading indicator easily.

How does one handle a retry? Setting the messageId again?

marc-scig: Yeah LoadStarted is a good call.

I didn't try reloading, but I think setting it again should be a reasonable way to go. We'll want to make sure it doesn't de-dupe though.

marc-scig: Dang, that doesn't work. Might need to add a call to broadcast the change to the message ID property once the page has already loaded.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/MessagePageRenderer.cs
marc-scig: gotcha

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/MessagePageRenderer.cs
```diff
@@ -0,0 +1,166 @@
...
+                var msw = MeasureSpec.MakeMeasureSpec(r - l, MeasureSpecMode.Exactly);
+                var msh = MeasureSpec.MakeMeasureSpec(b - t, MeasureSpecMode.Exactly);
+
+                nativeView.Measure(msw, msh);
+                nativeView.Layout(0, 0, r - l, b - t);
+            }
+        }
+
+        void RichPushInbox.IFetchMessagesCallback.OnFinished(bool success)
+        {
```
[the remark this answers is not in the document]
marc-scig: Should it?

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/Message Center/MessagePage.cs
marc-scig: Added property binding, as well as a LoadStarted event.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/MessagePageRenderer.cs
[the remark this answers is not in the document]
BrianBatchelder: Hmm. Seemed like the place to clean that up. But I guess the worse thing that could happen is that `Cancel()` could be called unnecessarily.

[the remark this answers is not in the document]
marc-scig: I can null it out, I just wasn't sure if you were implying that holding onto that reference would have nasty implications

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/MessagePageRenderer.cs
```diff
@@ -0,0 +1,161 @@
...
+            base.OnElementChanged(e);
+
+            if (e.OldElement != null || Element == null)
+            {
+                return;
+            }
+
+            messagePage = e.NewElement as MessagePage;
+            messageId = messagePage.MessageId;
+            messagePage.PropertyChanged += OnElementPropertyChanged;
```
marc-scig, line 124: For posterity: PageRenderer on iOS doesn't expose an OnElementPropertyChanged method to override, as it does on Android. So we have to make our own and subscribe to the event handler directly. On Android you just override the method.

andrewescutia:
> @BrianBatchelder re: forms dependencies, I wouldn't really call it a burden. Forms is how you write cross platform apps for Xamarin. If someone is using the NETStandard API they'll be using Forms already.

oh no. We are a case where we're using this library in a Xamarin Native app and currently have no dependencies to Forms :-(

PR #147, opened 2020-05-15
[Mobile-1487] Track Screen

benjamin-benoit:
### What do these changes do?
Add track screen tracking to Xamarin NetStandard library

To use screen tracking on xamarin

On emulator tests, I added on Android:
image
On iOS:
image
On the two sides app:
image

image
image

I also added TrackScreen method in AirshipBindings.NETStandart.Airship.cs, I don't know if it's good.


src/AirshipBindings.NETStandard/AirshipBindings.NETStandard/Airship.cs
```diff
@@ -83,6 +83,16 @@ public void AddCustomEvent(Analytics.CustomEvent customEvent)
             throw new NotImplementedException(BaitWithoutSwitchMessage);
         }
 
+        /// <summary>
+        /// track a screen for a specific app screen
+        /// </summary>
+        /// <param name="screen">The screen's identifier as an string
```
BrianBatchelder, line 89:
```suggestion
        /// <param name="screen">The screen's identifier.
```

PR #149, opened 2020-05-29
[MOBILE-590] .NetStandard Inbox events

Apekka:
### What do these changes do?
These changes expose in .Netstandard the method "displayMessage" to display a specific Inbox message,  and listeners for when the inbox is updated

To upgrade Xamarin

I edited the sample app with a button that display a specific message, verified that it displayed the message correctly.
And added listeners to see if I could saw logs telling that the inbox updated when a new message has come.


src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -82,7 +84,28 @@ public string NamedUser
...
+                onInboxUpdatedReceived += value;
+                UAirship.Shared().Inbox.RemoveListener(this);
+            }
+
+            remove
+            {
+                onInboxUpdatedReceived -= value;
+                if (onInboxUpdatedReceived == null)
+                {
+                    UAirship.Shared().Inbox.AddListener(null);
```
rlepinski, line 102: removeListener

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -266,5 +294,11 @@ public bool OnDeepLink(string deepLink)
             handler(deepLink);
             return true;
         }
+
+        public void OnInboxUpdated()
+        {
+            var handler = onInboxUpdatedReceived;
+            handler();
```
rlepinski, line 301: null check?

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -82,7 +84,28 @@ public string NamedUser
...
         }
-        
+
+        private InboxHandler onInboxUpdatedReceived;
+        public event InboxHandler OnInboxUpdatedReceived
+        {
+            add
+            {
+                onInboxUpdatedReceived += value;
+                UAirship.Shared().Inbox.RemoveListener(this);
```
rlepinski, line 94: You should hoo, up the listener once when the instance is created. Use the shared with a callback to do so: 
```
UAirship.Shared(() => {
                UAirship.Shared().Inbox.AddListener(this);
})
```

Then I dont think you need to override the event handler.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard/Airship.cs
```diff
@@ -102,6 +102,15 @@ public void DisplayMessageCenter()
...
 
+        /// <summary>
+        /// Displays a specific message.
+        /// </summary>
+        /// <param name="messageId">The identifier for the message to display.</param>
+        public void DisplayMessage(string messageId)
+        {
+            throw new NotImplementedException(BaitWithoutSwitchMessage);
+        }
+
```
rlepinski, line 113: Need to define the event listener in this file as well

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
rlepinski: iOS you can do the same. Instead of trying to use the delegate sign up for an NSNotificaationEvent - https://github.com/urbanairship/ios-library-dev/blob/master/Airship/AirshipMessageCenter/Source/Inbox/UAInboxMessageList.m#L14

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -82,7 +84,28 @@ public string NamedUser
                 }
             }
         }
-        
+
+        private InboxHandler onInboxUpdatedReceived;
+        public event InboxHandler OnInboxUpdatedReceived
```
rlepinski, line 89: I woold rename it to `OnMessageCenterUpdated`

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard/Airship.cs
```diff
@@ -7,6 +7,9 @@
 
 namespace UrbanAirship.NETStandard
 {
+    public delegate void DeepLinkHandler(string deepLink);
```
marc-scig, line 10: I think part of what has been causing problems is these custom delegate types. the fact that there were multiple declarations of InboxHandler before is probably why there were type errors at runtime.

We don't need to reinvent the wheel here. EventHandler is a perfectly idiomatic delegate type to use for events (see MessagePage for an example).

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -112,6 +113,25 @@ public string NamedUser
             }
         }
 
+        private InboxHandler onMessageCenterUpdated;
```
marc-scig, line 116: Let's either use EventHandler here or use an action delegate.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
marc-scig: The idiomatic way of invoking the handler while checking for null is this:

```
onMessageCenterUpdated?.Invoke()
```

If the hander is typed as EventHandler, then you invoke it like this:

```
onMessageCenterUpdated?.Invoke(this, EventArgs.Empty);
```

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard/Airship.cs
marc-scig: You can also use Action as a delegate, since it's already typed that way in the standard library. That would allow you to define events that use lambdas as the delegate part of the declaration, e.g.

```
public event Action OnMessageCenterUpdated;

...

OnMessageCenterUpdated += () => { 
    // do something
};
```

Or similarly for deep links:

```
public event Action<string> OnDeepLinkReceived
```

The advantage of this is that you don't need the sender or event args parameters that EventHandler uses, and you don't need to declare any extra types (such as custom subclasses of EventArgs) in order to pass custom event data. Which one you pick is up to you.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -111,6 +112,25 @@ public string NamedUser
             }
         }
 
+        private InboxHandler onMessageCenterUpdated;
```
marc-scig, line 115: Same general comments as in the android implementation

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
BrianBatchelder: A null check isn’t really needed, as the only way this call happens is when there is an event handler. However, it certainly doesn’t hurt to do the null check.

SampleApp/SampleApp.Android/SampleApp.Android.csproj
```diff
@@ -70,7 +70,7 @@
       <Version>4.6.0.726</Version>
     </PackageReference>
     <PackageReference Include="urbanairship.android.fcm">
-      <Version>12.2.3</Version>
```
rlepinski, line 73: Could you rebase your PR

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -112,6 +113,25 @@ public string NamedUser
...
+        public event InboxHandler OnMessageCenterUpdated
+        {
+            add
+            {
+                onMessageCenterUpdated += value;
+                MessageCenterClass.Shared().Inbox.AddListener(this);
```
rlepinski, line 122: If you call add more than once you will add the listener twice, causing duplicate events. You should only adad the listener once.

SampleApp/SampleApp/HomeViewController.xaml
```diff
@@ -9,6 +9,9 @@
                 <Image Source="urbanairship_logo"/>
             </StackLayout>
             <Button Text="{x:Static resources:AppResources.enable_push}" Clicked="OnButtonClicked" x:Name="enablePushButton"/>
+            <Button Text="{x:Static resources:AppResources.enable_push}" Clicked="MessageCenterBtnClicked" x:Name="displayMessageCenterButton"/>
+            <Button Text="{x:Static resources:AppResources.enable_push}" Clicked="DisplayMessageBtnClicked" x:Name="displayMessageButton"/>
```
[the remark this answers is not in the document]
Apekka: Not at all, my mistake

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
marc-scig: It's good practice to check for null when calling event handlers.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -112,6 +113,25 @@ public string NamedUser
             }
         }
 
+        private InboxHandler onMessageCenterUpdated;
+        public event InboxHandler OnMessageCenterUpdated
```
marc-scig, line 117: Try rewriting it like this

```
private addedInboxListener = false;
private EventHandler onMessageCenterUpdated;
public event EventHandler OnMessageCenterUpdated
{
    add
    {
        onMessageCenterUpdated += value;
        if (!addedInboxListener) {
            MessageCenterClass.Shared().Inbox.AddListener(this);
            addedInboxListener = true;
        }
    }

    remove
    {
        onMessageCenterUpdated -= value;
        if (onMessageCenterUpdated == null)
        {
            MessageCenterClass.Shared().Inbox.RemoveListener(this);
            addedInboxListener = false;
        }
    }
}
```

Then you don't need to define InboxHandler anywhere, and you can invoke `onMessageCenterUpdated` like this:

```
onMessageCenterUpdated?.Invoke(this, EventArgs.Empty);
```

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -112,6 +113,25 @@ public string NamedUser
...
+                onMessageCenterUpdated += value;
+                MessageCenterClass.Shared().Inbox.AddListener(this);
+            }
+
+            remove
+            {
+                onMessageCenterUpdated -= value;
+                if (onMessageCenterUpdated == null)
+                {
+                    MessageCenterClass.Shared().Inbox.RemoveListener(null);
```
marc-scig, line 130: we should be removing `this`, not `null`

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -111,6 +113,25 @@ public string NamedUser
...
+        public event InboxHandler OnMessageCenterUpdated
+        {
+            add
+            {
+                onMessageCenterUpdated += value;
+                NSNotificationCenter.DefaultCenter.AddObserver(aName: (NSString)"com.urbanairship.notification.message_list_updated", (notification) =>
+                {
+                    Console.WriteLine("Listener Set");
```
marc-scig, line 124: We should remove this line. We probably need a Logger class but that can be done in another ticket.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard/Airship.cs
marc-scig: Actually forget the deep link stuff for now, I'll make a separate ticket for it. Just focus on inbox events.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard/Airship.cs
marc-scig: You should add the declaration for `OnMessageCenterUpdated` to the IAirship interface so that compiler forces us to add it here. That's another reason for us to use EventHandler instead of custom delegates – you're not allowed to define delegates in an interface because interfaces can't have type declarations. Event declarations are OK though.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard/Airship.cs
marc-scig: Actually, don't worry about changing the deep link stuff. I'll make a separate ticket for that. Just focus on inbox events for now.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
Apekka: I should have listened to you earlier, that so much easier with the EventHandler !! Thank you !

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -112,6 +113,31 @@ public string NamedUser
...
         }
 
+        private Boolean addedInboxListener = false;
+        private EventHandler onMessageCenterUpdated;
+        public event EventHandler OnMessageCenterUpdated
+        {
+            add
+            {
+                onMessageCenterUpdated += value;
+                if (!addedInboxListener)
```
rlepinski, line 123: I think you should just add the listener in the constructor and not worry about removing it, or you could check `if (onMessageCenterUpdated == null)` before you add to see if you need to add the listener:

```
add
{
    if (onMessageCenterUpdated == null) {
       MessageCenterClass.Shared().Inbox.AddListener(this);
    }
    onMessageCenterUpdated += value;
}
```

But I think it would be safer to just add it at the beginning once, no reason to add/remove

Apekka: I could put it again at the initialization of Airship, but are you really sure nobody would like to remove the listeners? I'm kind of confused why a listener will be set at the beginning without control and then never removed

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -112,6 +112,28 @@ public string NamedUser
...
 
+        private EventHandler onMessageCenterUpdated;
+        public event EventHandler OnMessageCenterUpdated
+        {
+            add
+            {
+                if (onMessageCenterUpdated == null)
+                {
+                    MessageCenterClass.Shared().Inbox.AddListener(this);
+                    onMessageCenterUpdated += value;
```
rlepinski, line 123: you always have to add to the event handler

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -112,6 +112,28 @@ public string NamedUser
...
+                }
+            }
+
+            remove
+            {
+                if (onMessageCenterUpdated != null)
+                {
+                    MessageCenterClass.Shared().Inbox.RemoveListener(this);
+                    onMessageCenterUpdated -= value;
```
rlepinski, line 132: should subtract from it first, then check null

SampleApp/SampleApp/HomeViewController.xaml.cs
```diff
@@ -46,6 +46,17 @@ void OnButtonClicked(object sender, EventArgs e)
             }
         }
 
+        void MessageCenterBtnClicked(object sender, EventArgs e)
+        {
+            Airship.Instance.DisplayMessageCenter();
+         
+        }
+
+        void DisplayMessageBtnClicked(object sender, EventArgs e)
```
rlepinski, line 55: This is true for our messages, but its not always true. Lets remove this one

SampleApp/SampleApp/AppResources.resx
```diff
@@ -21,6 +21,12 @@
     <data name="enable_push" xml:space="preserve">
         <value>ENABLE PUSH</value>
     </data>
+    <data name="display_message_center" xml:space="preserve">
+        <value>DISPLAY MESSAGE CENTER</value>
+    </data>
+    <data name="display_message" xml:space="preserve">
```
rlepinski, line 27: Remove this

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -13,15 +13,33 @@ namespace UrbanAirship.NETStandard
...
+
+        private void Init()
+        {
+            //Adding Inbox updated listener
+            EventHandler handler = OnMessageCenterUpdated;
+            if (handler != null)
+            {
+                handler(this, EventArgs.Empty);
+            }
+            MessageCenterClass.Shared().Inbox.AddListener(this);
```
rlepinski, line 33: You should only add the listener here, the handler will always be empty at this point

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -377,5 +400,9 @@ public bool OnDeepLink(string deepLink)
             handler(deepLink);
             return true;
         }
+
+        public void OnInboxUpdated()
+        {
```
rlepinski, line 405: Need to add ```
//Adding Inbox updated listener
            EventHandler handler = OnMessageCenterUpdated;
            if (handler != null)
            {
                handler(this, EventArgs.Empty);
            }
```
here since this is the Airship callback when the inbox is updated.

File: src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs [hunk 22]
```diff
@@ -377,5 +395,15 @@ public bool OnDeepLink(string deepLink)
...
+            //Adding Inbox updated listener
+            EventHandler handler = OnMessageCenterUpdated;
```
marc-scig, line 402: An easier way to write this in one line:

`OnMessageCenterUpdated?.Invoke(this, EventArgs.Empty)`

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -206,6 +227,13 @@ public void DisplayMessageCenter()
             UAMessageCenter.Shared().Display();
         }
 
+        public void DisplayMessage(string messageId)
+        {
+            UAMessageCenter.Shared().DisplayMessage(messageId);
+        }
+
+
```
marc-scig, line 235: some whitespace to clean up here

SampleApp/SampleApp/HomeViewController.xaml.cs
```diff
@@ -46,6 +46,12 @@ void OnButtonClicked(object sender, EventArgs e)
             }
         }
 
+        void MessageCenterBtnClicked(object sender, EventArgs e)
+        {
+            Airship.Instance.DisplayMessageCenter();
+
```
marc-scig, line 52: whitespace

PR #151, opened 2020-06-02
[MOBILE-1621] fix syncVersion

marc-scig:
The`syncVersion` gradle task at the project root only deals with nuspecs and shared assembly info files. To sync everything correctly you need to run syncVersion in each of the subprojects, and those currently only depend on `build`. This has resulted in some confusion during development and review. I modified the script to address this. Tested manually by incrementing version numbers in airship.properties, running the task, and verifying the csproj files get updated correctly.


build.gradle
```diff
@@ -168,6 +168,21 @@ task syncVersion {
     }
 }
 
+// Sync subproject versions with the version declared in the airship.properties file.
+task syncVersion
+syncVersion.dependsOn('syncRootVersion')
+syncVersion.dependsOn(':src:AirshipBindings.Android.ADM:syncVersion',
```
BrianBatchelder, line 174: I'm not sure enumerating this is necessary. When `syncVersion` is run at the top level, it should run all subproject `syncVersion`s.

build.gradle
```diff
@@ -224,6 +239,18 @@ task clean(type: Delete) {
     }
 }
 
+clean.dependsOn(':src:AirshipBindings.Android.ADM:clean',
```
BrianBatchelder, line 242: I don't think this is necessary, like with `syncVersion`.

build.gradle
marc-scig: It wasn't doing that before I made this change.

build.gradle
marc-scig: Similarly, this was not the case before the change.

build.gradle
BrianBatchelder: That is totally strange. I thought the whole idea with the subprojects was that they all run the tasks. Then all you have to do is specify them once, in settings.gradle. @rlepinski ?

marc-scig:
I ended up reverting the changes to remove dependsOn logic. I believe we need these not to make sure all the sub tasks run, but so that they're guaranteed to run before other top level tasks that depend on them. I'm not sure if there's a better way of specifying that constraint, but this works. Otherwise you end up with packaging steps sometimes running before all the subproject build tasks have completed (for example).

PR #155, opened 2020-06-05
[MOBILE-1562] date attribute support

marc-scig:
This adds date attribute support to our NETStandard library. I tested it manually on iOS and Android with the new sample, which required a fair amount of work to get it to build. I'll call out those changes in the diff.

### DIFF
--- a/SampleApp/SampleApp.Android/SampleApp.Android.csproj
+++ b/SampleApp/SampleApp.Android/SampleApp.Android.csproj
@@ -70,7 +70,7 @@
       <Version>4.6.0.726</Version>
     </PackageReference>
     <PackageReference Include="urbanairship.android.fcm">
-      <Version>12.2.3</Version>
+      <Version>13.1.0</Version>
     </PackageReference>
   </ItemGroup>
   <ItemGroup>
--- a/SampleApp/SampleApp.Android/SampleAutopilot.cs
+++ b/SampleApp/SampleApp.Android/SampleAutopilot.cs
@@ -28,7 +28,7 @@ namespace Sample
             }
 
             UrbanAirshipListener airshipListener = new UrbanAirshipListener();
-            airship.PushManager.SetNotificationListener(airshipListener);
+            airship.PushManager.NotificationListener = airshipListener;
             airship.PushManager.AddPushListener(airshipListener);
             airship.PushManager.AddPushTokenListener(airshipListener);
             airship.Channel.AddChannelListener(airshipListener);
--- a/SampleApp/SampleApp.iOS/SampleApp.iOS.csproj
+++ b/SampleApp/SampleApp.iOS/SampleApp.iOS.csproj
@@ -79,13 +79,19 @@
     <Reference Include="System" />
     <Reference Include="System.Xml" />
     <Reference Include="System.Core" />
-    <Reference Include="Xamarin.iOS" />
     <Reference Include="System.Numerics" />
     <Reference Include="System.Numerics.Vectors" />
+    <Reference Include="Xamarin.iOS" />
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Xamarin.Forms" Version="4.5.0.657" />
     <PackageReference Include="Xamarin.Essentials" Version="1.5.3.1" />
+    <PackageReference Include="urbanairship.ios.core">
+      <Version>13.3.2</Version>
+    </PackageReference>
+    <PackageReference Include="urbanairship.ios.messagecenter">
+      <Version>13.3.2</Version>
+    </PackageReference>
   </ItemGroup>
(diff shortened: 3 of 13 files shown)


SampleApp/SampleApp.Android/SampleAutopilot.cs
```diff
@@ -28,7 +28,7 @@ public override void OnAirshipReady(UAirship airship)
             }
 
             UrbanAirshipListener airshipListener = new UrbanAirshipListener();
-            airship.PushManager.SetNotificationListener(airshipListener);
+            airship.PushManager.NotificationListener = airshipListener;
```
marc-scig, line 31: this was a compiler error

SampleApp/build.gradle
```diff
@@ -14,4 +14,4 @@ task build {
      }
 }
 }
-build.dependsOn(':src:AirshipBindings.NETStandard:build')
+build.dependsOn(':pkg')
```
marc-scig, line 17: This is to force the sample app to wait to build until packages are available. If you don't do that you can get into a state where you're unable to create local packages because SampleApp is failing to build when it can't find them.

SampleApp/nuget.config
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="Local" value="../build/" />
```
marc-scig, line 4: I think this will work in CI but we'll find out. If building SampleApp is going to be part of our gradle tasks and we want to be able to work with local packages then we can't depend on individual VS configurations.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -261,6 +261,21 @@ public List<MessageCenter.Message> InboxMessages
             }
         }
 
+        private NSDate FromDateTime(DateTime dateTime)
```
marc-scig, line 264: Note: I borrowed most of this logic from https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/unified/overview

src/AirshipBindings.iOS.Core/ApiDefinitions.cs
```diff
@@ -418,11 +418,6 @@ partial interface Constants
         [Obsolete("Deprecated – to be removed in SDK version 14.0. Please use UAFetchDeviceInfoActionDefaultRegistryName.")]
         NSString kUAFetchDeviceInfoActionDefaultRegistryName { get; }
 
-        // extern DEPRECATED_MSG_ATTRIBUTE( "Deprecated – to be removed in SDK version 14.0.") NSString *const kUAInteractionMCRAP
```
marc-scig, line 421: This was causing linker errors for SampleApp.iOS, due to the issue fixed in https://github.com/urbanairship/ios-library-dev/pull/1975

SampleApp/SampleApp.Android/SampleAutopilot.cs
BrianBatchelder: Why is the process of adding this listener different from the process of adding the other three? Shouldn't it be the same?

SampleApp/SampleApp.Android/SampleApp.Android.csproj
```diff
@@ -70,7 +70,7 @@
       <Version>4.6.0.726</Version>
     </PackageReference>
     <PackageReference Include="urbanairship.android.fcm">
-      <Version>12.2.3</Version>
+      <Version>13.1.0</Version>
```
BrianBatchelder, line 73: Android is now on 13.1.2

SampleApp/SampleApp.iOS/SampleApp.iOS.csproj
```diff
@@ -79,13 +79,19 @@
...
     <Reference Include="System.Core" />
-    <Reference Include="Xamarin.iOS" />
     <Reference Include="System.Numerics" />
     <Reference Include="System.Numerics.Vectors" />
+    <Reference Include="Xamarin.iOS" />
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Xamarin.Forms" Version="4.5.0.657" />
     <PackageReference Include="Xamarin.Essentials" Version="1.5.3.1" />
+    <PackageReference Include="urbanairship.ios.core">
```
BrianBatchelder, line 89: These two PackageReferences don't appear to be necessary. I got things working fine in APK's branch without them.

SampleApp/SampleApp.Android/SampleAutopilot.cs
marc-scig: Setters and getters are automatically bound to properties.

PR #156, opened 2020-06-17
Mobile-1560: Add channel registration listener

benjamin-benoit:
### What do these changes do?
Create/update event listener added to the .netstandard xamarin library.

Notify when the channel is created/updated.

iOS:
Airship.Instance.OnChannelCreation();
Android:
Airship.Instance.OnChannelUpdate += OnChannelUpdate;
static void OnChannelUpdate(string s)
{
Console.WriteLine("OnChannelUpdate");
Console.WriteLine(s);
}

iOS:
image
Android:
image

Android metadata xml fils updated.


src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -93,6 +96,48 @@ public string NamedUser
...
         }
 
+        private ChannelHandler onChannelCreation;
+        public event ChannelHandler OnChannelCreation
+        {
+            add
+            {
+                Console.WriteLine("onChannelCreationAdd");
+                onChannelCreation += value;
+                UAirship.Shared().Channel.AddChannelListener(this);
```
rlepinski, line 106: Pretty sure you cant follow this pattern since its more than one listener. You should add the listener once in the constructor or an init method otherwise you will send duplicate events for every added listener. Try adding two listeners to see what I mean.

benjamin-benoit: If I understand I have to add listener method (AddChannelListener) in an existing method like constructor or init method to don't have duplicate calls ?

rlepinski: Yeah, so in the constructor add the listener once, never remove it.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -377,5 +394,17 @@ public bool OnDeepLink(string deepLink)
...
+
+        public void OnChannelCreated(string channelId)
+        {
+            var handler = onChannelCreation;
+            handler(channelId);
+        }
+
+        public void OnChannelUpdated(string channelId)
+        {
+            var handler = onChannelUpdate;
```
rlepinski, line 406: i think you need to check if its null

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -14,13 +14,26 @@ namespace UrbanAirship.NETStandard
...
+            Airship instance = new Airship();
+            instance.Init();
+            return instance;
+        });
+
+        private void Init()
+        {
+            NSNotificationCenter.DefaultCenter.AddObserver(aName: (NSString)"com.urbanairship.channel.channel_created", (channelId) =>
+            {
+                Console.WriteLine(channelId);
```
rlepinski, line 28: need to call the handler here. What about channel updates? `com.urbanairship.channel.channel_updated`

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -10,21 +10,38 @@
...
         public static Airship Instance
         {
             get
             {
-                return sharedAirship;
+                return sharedAirship.Value;
             }
         }
 
+        public event ChannelHandler onChannelCreation;
```
marc-scig, line 41: instead of using a custom delegate, I'd prefer that we use EventHandler here. Make a class `ChannelEventArgs` that extends EventArgs, and then:

`public event EventHandler<ChannelEventArgs> onChannelCreation`

etc

https://docs.microsoft.com/en-us/dotnet/api/system.eventhandler-1?view=netcore-3.1

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
marc-scig: idiomatically: `onChannelUpdate?.Invoke(channelID)`

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
marc-scig: We should also make sure that these new events are in IAirship.cs, and using EventHandler<T> will make that easier since that type is already defined externally.

benjamin-benoit: Can you look if it's ok @marc-scig pls

benjamin-benoit: Okay I push an update but I get this: 'Airship' does not implement interface member 'IAirship.OnChannelCreation' and 'Airship' does not implement interface member 'IAirship.OnChannelUpdate'

marc-scig: @benjamin-benoit that's probably because there's a type mismatch. I'll point it out in the diff

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/IAirship.cs
```diff
@@ -73,5 +74,9 @@ List<MessageCenter.Message> InboxMessages
         Attributes.AttributeEditor EditChannelAttributes();
 
         Attributes.AttributeEditor EditNamedUserAttributes();
+
+        event EventHandler OnChannelCreation;
```
marc-scig, line 78: you want these to be `EventHandler<ChannelEventArgs>`. the ChannelEventArgs class will need to be defined outside the interface.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -377,5 +394,27 @@ public bool OnDeepLink(string deepLink)
...
+            OnChannelCreation?.Invoke(this, new ChannelEventArgs(channelId));
+        }
+
+        public void OnChannelUpdated(string channelId)
+        {
+            OnChannelUpdate?.Invoke(this, new ChannelEventArgs(channelId));
+        }
+    }
+
+    public class ChannelEventArgs: EventArgs
```
marc-scig, line 409: This class shouldn't be defined here, it needs to be accessible to both Android and iOS

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -14,13 +14,35 @@ namespace UrbanAirship.NETStandard
...
+            });
+
+            NSNotificationCenter.DefaultCenter.AddObserver(aName: (NSString)"com.urbanairship.channel.channel_updated", (channelId) =>
+            {
+                Console.WriteLine(channelId);
+            });
+        }
+
+        public event EventHandler OnChannelCreation;
```
marc-scig, line 37: these should be `EventHandler<ChannelEventArgs>`

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
rlepinski: You are still only logging the channel id instead of notifying the listeners.

```
            OnChannelCreation?.Invoke(this, new ChannelEventArgs(channelId));
```

But I am not actually what the event is structured, so you maya need to parse the event from the observer:

something like this:
```
(event) => { OnChannelCreation?.Invoke(this, new ChannelEventArgs(event["com.urbanairship.channel.identifier"]))
```

benjamin-benoit: I'm a pretty lost:
Are you telling me to do this on Android or iOS, because this line is one Android side:
`OnChannelCreation?.Invoke(this, new ChannelEventArgs(channelId));`

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -10,11 +10,14 @@
...
 namespace UrbanAirship.NETStandard
-{
+{ 
     public delegate void DeepLinkHandler(string deepLink);
 
-    public class Airship : Java.Lang.Object, IDeepLinkListener, IAirship, IInboxListener
+    public class Airship : Java.Lang.Object, IDeepLinkListener, IAirship, IInboxListener, UrbanAirship.Channel.IAirshipChannelListener
     {
+
+        public delegate void ChannelHandler(string channelId);
```
marc-scig, line 19: ChannelHandler is no loner needed

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
rlepinski: So you are taking a native event and pushing through as a c# event. On Android you add an event listener to Android SDK, then notify C# event listeners with  `OnChannelCreation?.Invoke(this, new ChannelEventArgs(channelId));`

On IOS, you are adding the event listener to the iOS SDK, but instead of notifying the C# event listeners you are just logging the event. You need to trigger the event handlers with `OnChannelCreation?.Invoke(this, new ChannelEventArgs(channelId))` like you do on Android.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard/Airship.cs
```diff
@@ -257,5 +257,37 @@ public Channel.TagGroupsEditor EditChannelTagGroups()
                 throw new NotImplementedException(BaitWithoutSwitchMessage);
             }
         }
+
+        /// <summary>
+        /// Add/remove the channel creation and update event listener.
+        /// </summary>
+        /// <value>The channel event listener.</value>
+        public delegate void ChannelHandler(string channelId);
```
marc-scig, line 265: ChannelHandler is no longer needed

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
marc-scig: The argument to the NSNotificationCenter callback Action is going to be an NSNotification. So it should look something like this:

```
NSNotificationCenter.DefaultCenter.AddObserver(aName: (NSString)"com.urbanairship.channel.channel_created", (NSNotification notification) =>
{
    string channelID = notification.UserInfo["com.urbanairship.channel.identifier"];
    OnChannelCreation?.Invoke(this, new ChannelEventArgs(channelId));
}
```

marc-scig: Or you can just get the channel from UAirship instead of parsing it out of the notification

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -22,6 +22,18 @@ public class Airship : UADeepLinkDelegate, IAirship
 
         private void Init()
         {
+            NSNotificationCenter.DefaultCenter.AddObserver(aName: (NSString)"com.urbanairship.channel.channel_created", (NSNotification notification) =>
+            {
+                string channelID = notification.UserInfo["com.urbanairship.channel.identifier"].ToString() ;
```
marc-scig, line 27: these ToString calls shouldn't be necessary. if it's a compiler error otherwise, I'd prefer that you solve it by casting, e.g.:

```
string channelID = (string) notification.UserInfo["com.urbanairship.channel.identifier"];
```

PR #158, opened 2020-06-20
[MOBILE-1010] Xamboni: Work with multiple modules

BrianBatchelder:
### What do these changes do?
Automates the generation of iOS bindings for multiple-modules.

As the number of modules in the SDK has grown, generating bindings for a release has gotten much harder.

Generated and compared bindings.

The path I chose to follow here was to do most of the automation in the gradle task. The task does the following:

1.  Use Carthage to get the iOS SDK. This only happens once per gradle execution, not once per module.
2. Run sourcekitten on the SDK. Sourcekitten needs to be pre-installed.
3. Run Xamboni from a submodule on the sourcekitten output. Xamboni is told to write the bindings to the correct folder.

There are other ways to do this. We could skip the submodule and instead ask the user to set an environment variable pointing at a cloned repo. We could automatically install sourcekitten if it isn't already installed. We could move the sourcekitten execution into Xamboni, rather than doing it as a separate step. I'm open to suggestions.

Note: There is also a [Xamboni PR](https://github.com/urbanairship/xamboni/pull/11) required for this work.


common-build.gradle
```diff
@@ -47,3 +47,133 @@ task syncVersionAndroid  {
     }
 }
 
+task generateIOSBindings {
+    doLast() {
+        // Only generate bindings for iOS projects
+        if ( !project.hasProperty('iosSdkModuleName') ) {
```
rlepinski, line 53: I feel like we should have another ios-common-build.gradle or something instead of inspecting the name

common-build.gradle
```diff
@@ -47,3 +47,133 @@ task syncVersionAndroid  {
...
+            return
+        }
+
+        println("Updating bindings for ${iosSdkModuleName} module.")
+
+        // clone iOS SDK at specified version, if it doesn't already exist
+        println("Using iOS SDK version ${airshipProperties.iosVersion}")
+        def tmpDir = System.getProperty("java.io.tmpdir")
+        def iosTempRepo = new File("${tmpDir}/ios-library-for-xamboni")
```
rlepinski, line 62: will this be shared across builds?

common-build.gradle
```diff
@@ -47,3 +47,133 @@ task syncVersionAndroid  {
...
+
+        boolean cloneRepo = false
+        if ( !iosTempRepo.exists() ) {
+            // cloned repo doesn't exist, so clone it
+            cloneRepo = true
+        } else {
+            // cloned repo exists, so check if it is the right version
+            def out = new ByteArrayOutputStream()
+            exec {
+                commandLine "git", "-C", "${iosTempRepo}", "tag", "--points-at", "HEAD"
```
rlepinski, line 73: This is ok, but I think we could use https://blog.gradle.org/introducing-source-dependencies to have gradle pull dependencies for us

common-build.gradle
```diff
@@ -47,3 +47,133 @@ task syncVersionAndroid  {
...
+                       "objective-c",
+                       "-isysroot", sdkPath,
+                       "-I", "${iosTempRepo}",
+                       "-I", "${iosTempRepo}/Airship/AirshipCore/Source/common",
+                       "-I", "${iosTempRepo}/Airship/AirshipCore/Source/ios",
+                       "-I", "${iosTempRepo}/${moduleSourcePath}",
+                       "-fmodules",
+                       "-D", "UA_USE_MODULE_IMPORT=0" ]
+
+        // manually add all folders in the module's Source directory, as I could not figure out how to get "-I ${iosTempRepo}/Airship/${MODULE_NAME}/Source/**"" to work
```
marc-scig, line 134: I"d maybe make this comment a little more concise and less less first person pronoun-y

common-build.gradle
```diff
@@ -47,3 +47,133 @@ task syncVersionAndroid  {
...
+            } else {
+                iosTempRepo.deleteDir()
+                cloneRepo = true
+            }
+        }
+
+        // actually clone the repo, if necessary
+        if (cloneRepo) {
+            exec {
+                commandLine "git", "clone", "--branch", "${airshipProperties.iosVersion}", "https://github.com/urbanairship/ios-library", "${iosTempRepo}"
```
marc-scig, line 88: is there any value in making this a separate task so that we have an easier time chaining it? I'm wondering what will happen if the clone step fails for some reason (network or otherwise). Would the gradle task exit cleanly if that were the case or would it result in a more obscure error further down?

common-build.gradle
```diff
@@ -47,3 +47,133 @@ task syncVersionAndroid  {
...
+        }
+
+        // Xamboni
+        println "Running Xamboni"
+
+        def xamboniTempRepo = new File("${tmpDir}/xamboni")
+        xamboniTempRepo.deleteOnExit()
+
+        boolean cloneXamboniRepo = false
+        if ( !xamboniTempRepo.exists() ) {
```
marc-scig, line 163: what if the cloned xamobni repo is out of date?

common-build.gradle
BrianBatchelder: Yes. It is deleted at the end of the build, or when the daemon exits, whichever happens last.

common-build.gradle
BrianBatchelder: That could happen, if the gradle daemon has been running for a long time. Maybe I should pull, also, or get Xamboni in a different way.

common-build.gradle
BrianBatchelder: Done.

common-build.gradle
BrianBatchelder: Done.

common-build.gradle
BrianBatchelder: I think this can be done, but it looks like a fair amount of work to learn how to do it and implement. I'm not sure it is worth it at this time.

common-build.gradle
BrianBatchelder: I added a `clean` task to run if there is any doubt as to the recency of the repo. `/gradlew clean generateIOSBindings` will remove the temporary iOS SDK and Xamboni repos, and then run `generateIOSBindings`, which will clone the repos anew.

ios-common-build.gradle
```diff
@@ -0,0 +1,168 @@
+task cloneIOSSDK {
+    doLast() {
+        // clone iOS SDK at specified version, if it doesn't already exist
+        println("Downloading iOS SDK version ${airshipProperties.iosVersion}")
+        def tmpDir = System.getProperty("java.io.tmpdir")
+        def iosTempRepo = new File("${tmpDir}/ios-library-for-xamboni")
```
marc-scig, line 6: This mostly seems like it should work, but it's a little weird to me that we're both cloning the repo directly into a temp directory with git, for this task, and using carthage for the frameworks. Is there a reason we need both? Could this potentially be simplified by using the carthage resources that the rest of the iOS gradle scripts depend on?

BrianBatchelder: That's is an excellent idea. I completely forgot that the Xamarin build uses Carthage. I'll take a look at it.

BrianBatchelder:
@marc-scig , @rlepinski : I used Marc's recommendation regarding the Carthage checkout, and re-organized things a bit to put the shared gradle files in a more logical place. Because Xamboni is small, and I changed the gradle so the clone only happens once for all of the generated bindings, I just clone it each time, so it is always fresh.

src/AirshipBindings.iOS.common/build.gradle
```diff
@@ -33,3 +33,42 @@ task build {
...
+        // Xamboni
+        def tmpDir = System.getProperty("java.io.tmpdir")
+        def xamboniTempRepo = new File("${tmpDir}/xamboni")
+        project.delete(files(xamboniTempRepo))
+
+        xamboniTempRepo.deleteOnExit()
+
+        println("Cloning Xamboni repo")
+        exec {
+            commandLine "git", "clone", "git@github.com:urbanairship/xamboni.git", "${xamboniTempRepo}"
```
marc-scig, line 58: Why are we still using git here? Does this task need to be removed?

src/AirshipBindings.iOS.common/build.gradle
```diff
@@ -33,3 +33,42 @@ task build {
...
+        }
+
+        checkoutXamboni.ext.repoPath = xamboniTempRepo
+    }
+}
+
+task cleanXamboniRepo(type: Delete) {
```
marc-scig, line 65: Does this need to be removed?

src/ios-common-build.gradle
```diff
@@ -0,0 +1,97 @@
...
+processIOSSDK.mustRunAfter('clean')
+
+task generateIOSBindings {
+    dependsOn 'processIOSSDK'
+    dependsOn ':src:AirshipBindings.iOS.common:checkoutXamboni'
+    doLast() {
+        println("Generating bindings for ${iosSdkModuleName} module.")
+
+        exec {
+            workingDir project(':src:AirshipBindings.iOS.common').checkoutXamboni.repoPath
```
marc-scig, line 93: I assume this is the wrong directory now

src/AirshipBindings.iOS.common/build.gradle
BrianBatchelder: I'm still cloning the Xamboni repo to get Xamboni. Did you have a different idea?

src/AirshipBindings.iOS.common/build.gradle
BrianBatchelder: This gives a way for devs to make sure they have the newest Xamboni, by running a clean build.

src/ios-common-build.gradle
BrianBatchelder: So `project(':src:AirshipBindings.iOS.common')` isn't specifying a directory, it is specifying a Gradle project, so I can get the `checkoutXamboni` task, and then the `repoPath` property on that task. That gives me that path to the temporary Xamboni repo.

src/AirshipBindings.iOS.common/build.gradle
marc-scig: Oh, sorry,  I misread that and thought it was cloning the iOS repo.

src/ios-common-build.gradle
marc-scig: ah got it.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "xamboni"]
+	path = xamboni
+	url = git@github.com:urbanairship/xamboni.git
```
rlepinski, line 3: Not sure how I feel about having a submodule. Do we have a different way of pulling it in, maybe with pip?

PR #160, opened 2020-06-29
[MOBILE-1722] standardize event handlers, fix iOS sample crash

marc-scig:
We've been slowly moving towards a style of events/delegates that uses `EventHandler<T>` instead of custom delegate types. This brings the deep link events up to date, and also fixes a few issues with the iOS part of the Sample app that I'll call out in the diff.

### DIFF
--- a/SampleApp/SampleApp.iOS/AppDelegate.cs
+++ b/SampleApp/SampleApp.iOS/AppDelegate.cs
@@ -14,7 +14,6 @@ using CoreFoundation;
 using Sample;
 
 using UrbanAirship;
-using Xamarin.Forms;
 
 namespace SampleApp.iOS
 {
@@ -28,6 +27,10 @@ namespace SampleApp.iOS
 
         public override bool FinishedLaunching(UIApplication app, NSDictionary options)
         {
+            global::Xamarin.Forms.Forms.Init();
+            LoadApplication(new App());
+            base.FinishedLaunching(app, options);
+
             this.FailIfSimulator();
 
             // Set log level for debugging config loading (optional)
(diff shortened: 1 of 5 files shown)


SampleApp/SampleApp.iOS/AppDelegate.cs
```diff
@@ -28,6 +27,10 @@ public partial class AppDelegate : global::Xamarin.Forms.Platform.iOS.FormsAppli
 
         public override bool FinishedLaunching(UIApplication app, NSDictionary options)
         {
+            global::Xamarin.Forms.Forms.Init();
```
marc-scig, line 30: When initializing a Forms app on the native side, the very first thing that has to happen is to initialize Forms, then load the application, then call super FinishedLaunching. Otherwise the application window won't exist when the rest of the the FinishedLaunching executes, and anything in this class that touches the window (such as `FailIfSimulator`) will crash.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -373,9 +373,11 @@ private void TagGroupHelper(List<Channel.TagGroupsEditor.TagOperation> payload,
 
         public bool OnDeepLink(string deepLink)
         {
-            var handler = onDeepLinkReceived;
-            handler(deepLink);
-            return true;
+            if (onDeepLinkReceived != null) {
```
marc-scig, line 376: handlers should always be checked for null. the easy way to do this in one line in c# is `handler?.Invoke(method)`, but in this case we also want to return a different result depending on whether we handled the link or not

PR #161, opened 2020-07-10
[MOBILE-1724] Add Unread attribute to the Message class

benjamin-benoit:
### What do these changes do?
Displays unread status to the Message center

See if the message is read or not

Unread status works and displayed

image


src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/MessageCenter/Message.cs
```diff
@@ -36,6 +36,12 @@ public partial class Message
         /// <value>The message expiration date.</value>
         public DateTime? ExpirationDate { get; }
 
+        /// <summary>
+        /// Gets the unread status boolean value.
+        /// </summary>
+        /// <value>The unread status.</value>
+        public bool Unread { get; }
```
rlepinski, line 43: IsRead is a better name for property

PR #164, opened 2020-07-23
MOBILE-1750 - Xamarin 13.1.0

marc-scig:
This PR is for the Xamarin 13.1.0 release. Since it's been a while, I updated the iOS SDK dependencies and bindings to 13.5.1 and fixed a few remaining problems I encountered while testing everything. Notably the deep link event code in .NETStandard was still problematic and causing crashes (I'll go into more detail in the diff). There were also problems with tab bar icons missing in the iOS native side of the SampleApp, due to inconsistencies in directory name casing. Directories with the .imageset extension have to have the same casing as the strings that refer to them in Forms ("home", "settings", etc) for then to show up correctly, and this is made more complicated by the fact that macOS is not case sensitive by default. 

Full disclosure, I used the refactored version of Xamboni still in review to generate the base content of these iOS bindings, as it was a good opportunity to demonstrate that it all still works OK. I manually fixed up the usual suspects as well as a few minor issues that I've made note of to address later. You will see that most of the changes in output are either formatting cleanup, non-binding changes to the way property declaration comments show up in sourcekitten, or the result of other recent Xamboni fixes that have already been reviewed. Please still take a close look at the new bindings and let me know if anything looks off.

Manual testing involved first getting the iOS and Android samples to run without crashing, which took a while, and then making sure that they behaved normally once they were running.


src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -113,25 +114,6 @@ public string NamedUser
             }
         }
 
-        private EventHandler<DeepLinkEventArgs> onDeepLinkReceived;
```
marc-scig, line 116: This section (same in iOS) was causing crashes. EventHandler is a delegate type, which is essentially a function pointer. If you declare one like this as an explicit backing delegate for an event, it will be initially as null, just like any other non-primitive variable. This is not C++, where declaring a variable invokes its class' default constructor. It's not that you can't do this, but rather that you have to check for null when using += to add listeners to it, such as below. I don't think this pattern of lazily adding native listeners buys us a whole lot, though, so I changed to be similar to the way we're dealing with Message Center events by adding listeners/NSNotification observers after takeOff.

src/AirshipBindings.iOS.Automation/ApiDefinitions.cs
```diff
@@ -487,7 +489,6 @@ interface UAActionAutomation
         // + (null_unspecified instancetype)shared;
         [Static]
         [Export("shared")]
-        [New]
```
BrianBatchelder, line 490: Why did the `[New]` annotation get removed? For the life of me I can't find any documentation on `[New]`. If you can drop me a URL, I'd appreciate it.

src/AirshipBindings.iOS.Core/ApiDefinitions.cs
```diff
@@ -1783,35 +1848,48 @@ interface UAJavaScriptEnvironment
         string Build ();
     }
 
+    // @interface UALocaleManager : NSObject /** * The currentLocale. Setting the currentLocale to nil reverts the locale to the current user’s chosen system locale. */ @property (nonatomic, strong, nullable) NSLocale *currentLocale; /** * Clear the locale. Reverts the locale to the current user’s chosen system locale. */ - (void)clearLocale; @end
```
BrianBatchelder, line 1851: I thought you had fixed these weird long comments, but I can't remember what the fix was.

src/AirshipBindings.iOS.MessageCenter/ApiDefinitions.cs
```diff
@@ -47,14 +46,17 @@ partial interface Constants
...
     interface UADefaultMessageCenterListViewController : IUITableViewDelegate, IUITableViewDataSource, IUIScrollViewDelegate
     {
-        // @property (nonatomic, strong, readwrite) UAMessageCenterStyle *_Nonnull style;
+        // @property (nonatomic, strong, readwrite) UAMessageCenterStyle *_Nonnull messageCenterStyle;
+        [Export("messageCenterStyle", ArgumentSemantic.Strong)]
+        UAMessageCenterStyle MessageCenterStyle { get; set; }
+
+        // @property(nonatomic, strong) UAMessageCenterStyle *style
         [Export("style", ArgumentSemantic.Strong)]
         UAMessageCenterStyle Style { get; set; }
```
BrianBatchelder, line 61: What happens when this is built for iOS 14? If there is a problem, is there anything we can do?

marc-scig: Oh you know that's a good question.

src/AirshipBindings.iOS.Core/ApiDefinitions.cs
marc-scig: Unfortunately these are coming from sourcekitten, so there's not a ton we can do unless we want to try to get clever with text processing

src/AirshipBindings.iOS.Automation/ApiDefinitions.cs
marc-scig: `[New]` is for constructors or factory methods. These are accessors, so they don't need it. You'll see that UAComponent doesn't have it in its Shared method, which these derive from.

marc-scig: Since these were added by hand, I think the `[New]` just kind of slipped in there

src/AirshipBindings.iOS.Core/ApiDefinitions.cs
BrianBatchelder: Let's not get clever.

src/AirshipBindings.iOS.Automation/ApiDefinitions.cs
marc-scig: https://stackoverflow.com/questions/6576206/what-is-the-difference-between-the-override-and-new-keywords-in-c/6576212#6576212

marc-scig: It makes a warning go away, but otherwise doesn't do anyting

src/AirshipBindings.iOS.MessageCenter/ApiDefinitions.cs
marc-scig: I'm going to rebind these to all point to the new properties, which should cover all our bases since they will always be present and will always work.

src/AirshipBindings.iOS.Automation/ApiDefinitions.cs
marc-scig: I'll add them back in. We'll need to update Xamboni to use them when reconstructing inherited factory methods.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -29,12 +26,16 @@ private void Init()
             
             //Adding Inbox updated listener
             MessageCenterClass.Shared().Inbox.AddListener(this);
+
+            UAirship.Shared().DeepLinkListener = this;
```
rlepinski, line 30: This is problematic since it's not a single listener. Now by including the .net framework it takes over the listener instead of only doing it if you use this functionality. This could break other apps.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -29,12 +26,44 @@ private void Init()
...
```
BrianBatchelder, line 30: Do you need to remove this line, now that you are handling it below?

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -43,12 +42,46 @@ private void Init()
                     handler(this, EventArgs.Empty);
                 }
             });
+
+            NSNotificationCenter.DefaultCenter.AddObserver(aName: (NSString)"com.urbanairship.airship_ready", (notification) =>
+            {
+                UAirship.Shared().WeakDeepLinkDelegate = this;
+            });
+
```
BrianBatchelder, line 50: Do you need to remove this line, now that you are handling it below?

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
marc-scig: oops

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
marc-scig: Thanks for noticing that. Fix on the way.

marc-scig:
This isn't ready to merge, I need to make some SampleApp changes since I misdiagnosed the event crashes. We will probably be releasing Monday

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
marc-scig: I misunderstood the source of the crash. Disregard the above.

PR #167, opened 2020-09-04
[MOBILE-1890] Add a way to mark message center messages read and deleted

Apekka:
### What do these changes do?
These changes add methods "MarkAsRead" and "DeleteMessage" to Xamarin NETStandard

Because customers need this

By adding a button "mark as read" and "delete" in the sample app, and checking that it's working well.

### DIFF
--- a/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
+++ b/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
@@ -216,6 +216,16 @@ namespace UrbanAirship.NETStandard
             MessageCenterClass.Shared().ShowMessageCenter(messageId);
         }
 
+        public void MarkMessageRead(string messageId)
+        {
+            MessageCenterClass.Shared().Inbox.GetMessage(messageId).MarkRead();
+        }
+
+        public void DeleteMessage(string messageId)
+        {
+            MessageCenterClass.Shared().Inbox.GetMessage(messageId).Delete();
+        }
+
         public int MessageCenterUnreadCount
         {
             get
--- a/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
+++ b/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
@@ -218,12 +218,20 @@ namespace UrbanAirship.NETStandard
 
         public void MarkMessageRead(string messageId)
         {
-            MessageCenterClass.Shared().Inbox.GetMessage(messageId).MarkRead();
+            var toRead = new List<String>
+            {
+                messageId
+            };
+            MessageCenterClass.Shared().Inbox.MarkMessagesRead(toRead);
         }
 
         public void DeleteMessage(string messageId)
         {
-            MessageCenterClass.Shared().Inbox.GetMessage(messageId).Delete();
+            var toDelete = new List<String>
+            {
+                messageId
+            };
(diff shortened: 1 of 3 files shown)


src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -216,6 +216,16 @@ public void DisplayMessage(string messageId)
             MessageCenterClass.Shared().ShowMessageCenter(messageId);
         }
 
+        public void MarkMessageRead(string messageId)
+        {
+            MessageCenterClass.Shared().Inbox.GetMessage(messageId).MarkRead();
```
rlepinski, line 221: I think you will face NPE if the message is not available. Try:
`MessageCenterClass.Shared().Inbox.MarkMessagesRead/DeleteMessages`

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -244,6 +244,34 @@ public void DisplayMessage(string messageId)
...
+        {
+            var toDelete = new UAInboxMessage[1];
+            var messages = UAMessageCenter.Shared().MessageList.Messages;
+            foreach (var message in messages)
+            {
+                if (message.MessageID == messageId)
+                {
+                    toDelete[0] = message;
+                    UAMessageCenter.Shared().MessageList.MarkMessagesDeleted(toDelete, null);
+                    Console.WriteLine("Message deleted !");
```
rlepinski, line 270: probably want to remove the logs

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -244,6 +244,34 @@ public void DisplayMessage(string messageId)
...
+                    message.MarkMessageRead(null);
+                    Console.WriteLine("Message read !");
+                }
+            }
+        }
+
+        public void DeleteMessage(string messageId)
+        {
+            var toDelete = new UAInboxMessage[1];
+            var messages = UAMessageCenter.Shared().MessageList.Messages;
```
rlepinski, line 263: Should be able to get the message with messageForID method

Apekka: Looks like the method messageForID is not exposed, I found it in ApiDefinitions.cs (cf this screenshot), but couldn't retrieve it in the code, I don't know why
image

marc-scig: `var message = UAMessageCenter.Shared().MessageList.Message(messageId);`

marc-scig (commented):
I think the iOS implementation can be simplified. See my comment on DeleteMessage

Ulrico972 (approved):

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -244,6 +244,25 @@ public void DisplayMessage(string messageId)
             UAMessageCenter.Shared().DisplayMessage(messageId);
         }
 
+        public void MarkMessageRead(string messageId)
+        {
+            var messages = UAMessageCenter.Shared().MessageList.Messages;
```
rlepinski, line 249: i believe you can do the same thing as you are doing for delete

marc-scig (approved):

PR #168, opened 2020-09-22
[MOBILE-1898] Updating Xamarin to Android 14.0.0 and iOS 14.1.1

Apekka:
### What do these changes do?
Updating Xamarin to version 14.0.0, supporting Android SDK 14.0.0 & iOS SDK 14.1.1

To stay up to date

Testing sample app


marc-scig (commented):
If the packages are not building for you locally then it's not actually tested. You'll need the changes in the `wcarthage` branch in order to get the carthage build failures to go away. We also need to generate new bindings for the iOS SDK so that any new or changed APIs (such as the URL allow list stuff mentioned in the CHANGELOG) are available.  You can get that process started running the `generateIOSBindings` task in gradle for each of the modules.

marc-scig:
if you haven't already you should also install pyenv and make your global python version 3.x so that xamobni will work

marc-scig:
Packages should be building locally, and iOS changes ready for review now. I think we still need to do some work on Android to support Room (probably need to tweak the Android X package dependencies).

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Airship Xamarin Changelog
 
+## Version 14.0.0 - Sept 28, 2020
+Major release to support iOS SDK 14.1.2 & Android SDK 14.0.1.
+
+### Changes
```
rlepinski, line 6: We also added some methods on deleting/reading messages and fixes the channel update listener.

rlepinski: @oristanovic has docs that need https://urbanairship.atlassian.net/browse/MOBILE-1924 in as well. Looks like a small lift, can we get it in real fast for the release?

rlepinski (approved):

rlepinski (approved):

PR #169, opened 2020-09-23
[MOBILE-1887] iOS module init

marc-scig:
We've had a few reports that IAA doesn't work in Xamarin, and it turns out that the module is being stripped out when it's not referenced in code. I've tried various things from linker settings to Preserve annotations, but the one thing that seems to work is adding a dummy  static "Init" method that can be called to force the compiler to include the module. This is a potential problem for any module aside from AirshpCore, which you already have to reference in order to call takeOff.

Since the NETStandard library only directly references Core and Message Center, I modified it to initialize Automation and Extended Actions, so that NETStandard users don't have to bother about this. Native apps will need to call the appropriate init methods for the modules they're using, which will need clear documentation.

Tested this by building the packages locally, running our SampleApp and verifying that an automation I sent to myself was displayed.

### DIFF
--- a/SampleApp/SampleApp.iOS/SampleApp.iOS.csproj
+++ b/SampleApp/SampleApp.iOS/SampleApp.iOS.csproj
@@ -87,7 +87,7 @@
     <PackageReference Include="Xamarin.Forms" Version="4.5.0.657" />
     <PackageReference Include="Xamarin.Essentials" Version="1.5.3.1" />
     <PackageReference Include="urbanairship.ios">
-      <Version>13.5.4</Version>
+      <Version>13.5.2</Version>
     </PackageReference>
   </ItemGroup>
   <Import Project="$(MSBuildExtensionsPath)\Xamarin\iOS\Xamarin.iOS.CSharp.targets" />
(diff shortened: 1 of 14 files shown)


SampleApp/SampleApp.iOS/SampleApp.iOS.csproj
```diff
@@ -86,11 +86,8 @@
   <ItemGroup>
     <PackageReference Include="Xamarin.Forms" Version="4.5.0.657" />
     <PackageReference Include="Xamarin.Essentials" Version="1.5.3.1" />
-    <PackageReference Include="urbanairship.ios.core">
-      <Version>13.5.2</Version>
-    </PackageReference>
-    <PackageReference Include="urbanairship.ios.messagecenter">
-      <Version>13.5.2</Version>
+    <PackageReference Include="urbanairship.ios">
```
rlepinski, line 89: whats the dealio above?

rlepinski (approved):
Assuming version stuff is fine

SampleApp/SampleApp.iOS/SampleApp.iOS.csproj
marc-scig: Oh, I should have mentioned this in the description. one of the problems I noticed with the sample app was that it didn't include automation to begin with, so if someone was copying its structure they might end up with an app that won't be able to get IAA. Since we don't provide direct IAA APIs in the .NETStandard library I figured the best thing would be to just use the umbrella package.

marc-scig: but yeah the version is wrong right now since it came from the local build

PR #173, opened 2020-10-15
[MOBILE-1932] Add Xamarin forms message center listener

oristanovic:
### What do these changes do?
Add a message center display listener. If set, displayMessageCenter and displayMessage route to the built in message center. If not, the default message center is displayed.

Allow for the custom Xamarin Forms MC to be used instead of the OOTB one.

Ran sample apps and verified that the behavior is as expected.

### DIFF
--- a/SampleApp/SampleApp/App.xaml.cs
+++ b/SampleApp/SampleApp/App.xaml.cs
@@ -44,7 +44,7 @@ namespace SampleApp
         {
             Console.WriteLine("onMessageCenterUpdated");
         }
-        static void OnMessageCenterDisplay(object sender, EventArgs e)
+        static void OnMessageCenterDisplay(object sender, MessageCenterEventArgs e)
         {
             TabbedPage originalRootPage = (TabbedPage)App.Current.MainPage.Navigation.NavigationStack.Last();
 
--- a/SampleApp/SampleApp/App.xaml.cs
+++ b/SampleApp/SampleApp/App.xaml.cs
@@ -7,6 +7,7 @@ using Xamarin.Forms;
 using Xamarin.Forms.Xaml;
 
 using UrbanAirship.NETStandard;
+using System.Linq;
 
 namespace SampleApp
 {
@@ -23,6 +24,7 @@ namespace SampleApp
         {
             Airship.Instance.OnDeepLinkReceived += OnDeepLinkReceived;
             Airship.Instance.OnMessageCenterUpdated += OnMessageCenterUpdated;
+            Airship.Instance.OnMessageCenterDisplay += OnMessageCenterDisplay;
         }
 
         protected override void OnSleep()
@@ -41,6 +43,12 @@ namespace SampleApp
         static void OnMessageCenterUpdated(object sender, EventArgs e)
         {
             Console.WriteLine("onMessageCenterUpdated");
+        }
+        static void OnMessageCenterDisplay(object sender, EventArgs e)
+        {
+            TabbedPage originalRootPage = (TabbedPage)App.Current.MainPage.Navigation.NavigationStack.Last();
+
+            originalRootPage.CurrentPage = originalRootPage.Children[1];
         }
     }
 }
--- a/SampleApp/SampleApp/App.xaml.cs
+++ b/SampleApp/SampleApp/App.xaml.cs
@@ -7,6 +7,7 @@ using Xamarin.Forms;
 using Xamarin.Forms.Xaml;
 
 using UrbanAirship.NETStandard;
+using UrbanAirship.NETStandard.MessageCenter;
 using System.Linq;
 
 namespace SampleApp
@@ -49,6 +50,13 @@ namespace SampleApp
             TabbedPage originalRootPage = (TabbedPage)App.Current.MainPage.Navigation.NavigationStack.Last();
 
             originalRootPage.CurrentPage = originalRootPage.Children[1];
+
+            if (e.MessageId != null)
+            {
+                var messagePage = new MessagePage();
+                messagePage.MessageId = e.MessageId;
+                originalRootPage.Navigation.PushAsync(messagePage);
+            }
         }
     }
 }
--- a/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/IAirship.cs
+++ b/src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/IAirship.cs
@@ -28,6 +28,15 @@ namespace UrbanAirship.NETStandard
         }
     }
 
+    public class MessageCenterEventArgs : EventArgs
+    {
+        public string MessageId { get; internal set; }
+        public MessageCenterEventArgs(string messageId = null)
+        {
+            MessageId = messageId;
+        }
+    }
+
     public interface IAirship
     {
         bool UserNotificationsEnabled
@@ -66,6 +75,8 @@ namespace UrbanAirship.NETStandard
 
         event EventHandler<DeepLinkEventArgs> OnDeepLinkReceived;
 
+        event EventHandler<MessageCenterEventArgs> OnMessageCenterDisplay;
+
         Channel.TagEditor EditDeviceTags();
 
         void AddCustomEvent(Analytics.CustomEvent customEvent);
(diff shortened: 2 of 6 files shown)


SampleApp/SampleApp/App.xaml.cs
```diff
@@ -42,5 +44,11 @@ static void OnMessageCenterUpdated(object sender, EventArgs e)
         {
             Console.WriteLine("onMessageCenterUpdated");
         }
+        static void OnMessageCenterDisplay(object sender, EventArgs e)
```
rlepinski, line 47: Should this be MessageCenterEventArgs ?

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/IAirship.cs
```diff
@@ -28,6 +28,15 @@ public DeepLinkEventArgs(string deepLink)
         }
     }
 
+    public class MessageCenterEventArgs : EventArgs
+    {
+        public string MessageId { get; internal set; }
```
rlepinski, line 33: Do we need to specify that this is optional?

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -211,12 +213,26 @@ public void AssociateIdentifier(string key, string identifier)
...
         public void DisplayMessageCenter()
         {
-            MessageCenterClass.Shared().ShowMessageCenter();
+            if (OnMessageCenterDisplay != null)
+            {
+                OnMessageCenterDisplay.Invoke(this, new MessageCenterEventArgs());
+            }
+            else
+            {
+                MessageCenterClass.Shared().ShowMessageCenter();
```
rlepinski, line 222: You should only have to call this here. If you set the message center listener on the airship sdk, it will forward to the right place if set - https://docs.airship.com/platform/android/message-center/#overriding-the-default-display

rlepinski: The approach should be, if you have a netstandard listener, set the SDK listener, otherwise clear it.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Android/Airship.cs
```diff
@@ -211,12 +213,26 @@ public void AssociateIdentifier(string key, string identifier)
...
         public void DisplayMessage(string messageId)
         {
-            MessageCenterClass.Shared().ShowMessageCenter(messageId);
+            if (OnMessageCenterDisplay != null)
+            {
+                OnMessageCenterDisplay.Invoke(this, new MessageCenterEventArgs(messageId));
+            }
+            else
+            {
+                MessageCenterClass.Shared().ShowMessageCenter(messageId);
```
rlepinski, line 234: same on this one

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -247,12 +249,27 @@ public void AssociateIdentifier(string key, string identifier)
...
         public void DisplayMessage(string messageId)
         {
-            UAMessageCenter.Shared().DisplayMessage(messageId);
+            if (OnMessageCenterDisplay != null)
+            {
+                OnMessageCenterDisplay.Invoke(this, new MessageCenterEventArgs(messageId));
+            }
+            else
+            {
+                UAMessageCenter.Shared().DisplayMessage(messageId);
```
rlepinski, line 271: Same idea on iOS except its a delegate that you will have to manage. I think this follows same pattern as the deep link listener.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/IAirship.cs
oristanovic: Seems that setting a default value makes it optional, not sure if there's another way to do it?

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -471,6 +490,19 @@ override public void ReceivedDeepLink(NSUrl url, Action completionHandler)
...
+            onMessageCenterDisplay?.Invoke(this, new MessageCenterEventArgs(messageID));
+        }
+
+        public void DisplayMessageCenterAnimated(bool animated)
+        {
+            onMessageCenterDisplay?.Invoke(this, new MessageCenterEventArgs());
+        }
+
+        public void DismissMessageCenterAnimated(bool animated)
+        { }
```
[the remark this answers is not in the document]
rlepinski: When does this get called in iOS? Maybe its fine in a cross platform implementation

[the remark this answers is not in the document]
Ulrico972: I don't think this is used in iOS. I feel like we don't need this, but I have to say I still have some difficulties understanding how EventHandler and EventArgs work.

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/IAirship.cs
Ulrico972: We can't mark strings as nullable, that would require Xamarin 8.0 language.

oristanovic: @marc-scig I would still need to pass it as an argument in that case. Do we want it to be optional or nullable?

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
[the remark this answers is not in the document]
oristanovic: Not sure what the implementation would be?

SampleApp/SampleApp/App.xaml.cs
```diff
@@ -42,5 +44,11 @@ static void OnMessageCenterUpdated(object sender, EventArgs e)
         {
             Console.WriteLine("onMessageCenterUpdated");
         }
+        static void OnMessageCenterDisplay(object sender, MessageCenterEventArgs e)
+        {
+            TabbedPage originalRootPage = (TabbedPage)App.Current.MainPage.Navigation.NavigationStack.Last();
+
+            originalRootPage.CurrentPage = originalRootPage.Children[1];
```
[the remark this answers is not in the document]
oristanovic: yep

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/IAirship.cs
rlepinski: Optional seems fine to me

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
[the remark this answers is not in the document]
rlepinski: We can skip it for now

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.Abstractions/IAirship.cs
marc-scig: Yeah I'd probably go with optional here

src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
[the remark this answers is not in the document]
marc-scig: agreed

marc-scig (approved):
+1 with the optional parameter thng

rlepinski (approved):

PR #175, opened 2020-12-24
MOBILE-2031 Update Xamarin with latest Android SDK 14.1 and iOS SDK 14.2

marc-scig:
I had to update the android dependencies a bit to get everything to work, but otherwise it's not a huge change. Tested manually on iOS Simulator and Android emulators to make sure the app runs and doesn't have any runtime crashes.

### DIFF
--- a/UrbanAirship.Android.FCM.nuspec
+++ b/UrbanAirship.Android.FCM.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
    <metadata>
       <id>urbanairship.android.fcm</id>
-      <version>14.0.1</version>
+      <version>14.1.0</version>
       <title>Airship Android SDK - FCM Push Provider</title>
       <authors>Urban Airship, Inc.</authors>
       <projectUrl>https://github.com/urbanairship/urbanairship-xamarin</projectUrl>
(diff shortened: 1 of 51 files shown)


UrbanAirship.Android.FCM.nuspec
```diff
@@ -11,9 +11,10 @@
 
       <dependencies>
          <group targetFramework="MonoAndroid4.1">
-            <dependency id="Xamarin.Firebase.Messaging" version="71.1740.0" />
-            <dependency id="Xamarin.GooglePlayServices.Base" version="71.1610.0" />
-            <dependency id="urbanairship.android.core" version="14.0.1"/>
+            <dependency id="Xamarin.Firebase.Messaging" version="120.3.0" />
+            <dependency id="Xamarin.GooglePlayServices.Base" version="117.1.0" />
+            <dependency id="Xamarin.Google.Dagger" version="2.27.0" />
```
marc-scig, line 16: I had to add this to resolve a runtime crash during Firebase initialization. It's a transitive dependency, but wasn't getting pulled in for some reason.

src/AirshipBindings.Android.Core/Transforms/Metadata.xml
```diff
@@ -25,7 +25,8 @@
   <remove-node path="/api/package[@name='com.urbanairship.job']/class[@name='Job']" />
   <remove-node path="/api/package[@name='com.urbanairship.job']/class[@name='JobDispatcher']" />
   <remove-node path="/api/package[@name='com.urbanairship.job']/class[@name='AndroidJobScheduler']" />
-	
+  <remove-node path="/api/package[@name='com.urbanairship.channel']/class[@name='NamedUser']/method[@name='addNamedUserListener']" />
```
marc-scig, line 28: The bindings generator turns listener adders into C# events, and the generated event was called NamedUser which conflicted with the class. Removing it was the easiest solution, since it's for internal use and we don't need to expose it anyway.

rlepinski (approved):

PR #179, opened 2021-05-18
[MOBILE-2240] iOS SDK 14.4.1, chat

marc-scig:
This adds bindings for AirshipChat and bumps the rest of the iOS SDK bindings to 14.4.1. I also took the liberty of changing the build scripts to use xcframeworks downloaded from GitHub instead of building the frameworks directly from Carthage. This cuts down on build times significantly – on my machine `gradlew clean build pkg` takes about 3 to 4 minutes now. However it turns out xcframeworks support is still somewhat incomplete in Xamarin, so while building the packages is straightforward, you have to do a few hacks with .targets files, nuspecs and csproj settings to make the packages usable. I'll call out some examples in the diff.

You may also notice changes to cross platform and android version numbers. These were already in `airship.properties`, my guess is that in the last release the versions were not synced properly before publishing.

Because AirshipChat is our first public-facing native Swift module, bindings are tricky right now. So far I have not gotten sourcekitten to play well with xcframeworks, and so I kept the carthage checkout task but I'm just using the source code to generate bindings, without actually building the frameworks. That works for Objective-C source code but not for Swift, since there's no bridging header. For this PR I used Sharpie to get me part of the way there, and manually finished the rest. What we'll probably want to do eventually is to modify Xamboni to work with Swift source code directly (sourcekitten is probably better at this anyway) but I've saved that for a follow up ticket.

Simulator Screen Shot - iPod touch (7th generation) - 2021-05-18 at 14 48 15


SampleApp/SampleApp.iOS/SampleContentExtension/SampleContentExtension.csproj
```diff
@@ -107,4 +107,5 @@
     </Compile>
   </ItemGroup>
   <Import Project="$(MSBuildExtensionsPath)\Xamarin\iOS\Xamarin.iOS.AppExtension.CSharp.targets" />
+  <Import Project="..\..\packages\urbanairship.ios.notificationcontentextension.14.2.3\build\urbanairship.ios.notificationcontentextension.targets" Condition="Exists('..\..\packages\urbanairship.ios.notificationcontentextension.14.2.3\build\urbanairship.ios.notificationcontentextension.targets')" />
```
marc-scig, line 110: The user doesn't need to do anything special to make this happen, it will automatically get imported along with the package

UrbanAirship.iOS.Accengage.nuspec
```diff
@@ -15,5 +15,7 @@
 
    <files>
      <file src="src/AirshipBindings.iOS.Accengage/bin/release/AirshipBindings.iOS.Accengage.dll" target="lib/Xamarin.iOS10" />
+     <file src="src/AirshipBindings.iOS.Accengage/bin/Release/AirshipBindings.iOS.Accengage.resources/**" target="content/AirshipBindings.iOS.Accengage.resources" />
```
marc-scig, line 18: This line copies over the "resources" directory to the target. The resources directory contains the xcframework as well as a manifest file (autogenerated) that tells VS how to actually link the xcframework, which contains multiple architecture slices that currently confuse the linker without further instruction. We may be able to remove this at some point in the future.

UrbanAirship.iOS.Accengage.nuspec
```diff
@@ -15,5 +15,7 @@
...
+     <file src="src/AirshipBindings.iOS.Accengage/urbanairship.ios.accengage.targets" target="build/urbanairship.ios.accengage.targets" />
```
marc-scig, line 19: This copies the .targets file, which is necessary to get VS to link against the xcframework copied in the resources directory

src/AirshipBindings.iOS.Accengage/AirshipBindings.iOS.Accengage.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
+    <NoBindingEmbedding>true</NoBindingEmbedding>
```
marc-scig, line 4: This is part of the workaround to support xcframeworks

src/AirshipBindings.iOS.Accengage/AirshipBindings.iOS.Accengage.csproj
```diff
@@ -34,6 +35,11 @@
     <Reference Include="System" />
     <Reference Include="Xamarin.iOS" />
   </ItemGroup>
+  <ItemGroup>
+    <None Include="urbanairship.ios.accengage.targets">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
marc-scig, line 40: this makes sure our custom .targets file ends up in the output

src/AirshipBindings.iOS.Accengage/urbanairship.ios.accengage.targets
```diff
@@ -0,0 +1,14 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <Target Name="AirshipAccengageBeforeCompile" AfterTargets="BeforeCompile">
+        <ItemGroup>
+            <BindingResources Include="$(MSBuildThisFileDirectory)../content/AirshipBindings.iOS.Accengage.resources/**/*.*" />
+        </ItemGroup>
+        <Copy SourceFiles="@(BindingResources)" DestinationFolder="$(TargetDir)/AirshipBindings.iOS.Accengage.resources/%(RecursiveDir)" ContinueOnError="false" />
+        <ItemGroup>
+            <NativeReference Include="$(TargetDir)/AirshipBindings.iOS.Accengage.resources\AirshipAccengage.xcframework">
```
marc-scig, line 8: This makes VS link against the framework directly (instead of implicitly through the DLL) if someone imports the nupkg into their project

src/AirshipBindings.iOS.Automation/ApiDefinitions.cs
```diff
@@ -2151,6 +2168,43 @@ interface UAScheduleAudience
         UAScheduleAudience Audience (Action<UAScheduleAudienceBuilder> builderBlock);
     }
 
+    // @interface UAScheduleDeferredData : NSObject
+    [BaseType(typeof(NSObject))]
+    interface UAScheduleDeferredData
```
rlepinski, line 2173: can we hide this? it should be hidden in the main sdk

rlepinski (approved):
Looks good

PR #180, opened 2021-05-22
MOBILE-2239 Android chat bindings

marc-scig:
This was challenging but it mostly came down to dependencies. There are bugs in the xamarin bindings of work manager, which make it difficult to diagnose exceptions thrown within worker threads. We were missing OK HTTP and the Xamarin.KotlinX.Coroutines.Android package. Since we're using a prebuilt aar for chat at the moment, I commented out some stuff in the gradle files and .gitignore so that we can check in the aar and not have the gradle task wipe the jars directory or download chat aars from maven.

I also went ahead and updated the cross platform version and CHANGELOG in this PR, so we can expedite the release.

Screenshot_1621641988


jyaganeh (approved):
:+1:

PR #190, opened 2022-01-05
Reduce nuget path length

khmMouna:
Use the format **UAModuleName.resources** instead of **AirshipBindings.iOS.Automation.resources** to reduce the nuget length path. As suggest @marc-scig I implemented the fix from this https://github.com/xamarin/xamarin-macios/issues/10819


UrbanAirship.iOS.Basement.nuspec
```diff
@@ -15,7 +15,7 @@
 
    <files>
      <file src="src/AirshipBindings.iOS.Basement/bin/release/AirshipBindings.iOS.Basement.dll" target="lib/Xamarin.iOS10" />
-     <file src="src/AirshipBindings.iOS.Basement/bin/Release/AirshipBindings.iOS.Basement.resources/**" target="content/AirshipBindings.iOS.Basement.resources" />
+     <file src="src/AirshipBindings.iOS.Basement/bin/Release/AirshipBindings.iOS.Basement.resources/**" target="content/Basement.resources" />
```
rlepinski, line 18: do we risk having conflicts with other libs? should we add a `ua` prefix?

khmMouna: I am not sure but we can add the prefix to avoid any conflit just in case.

marc-scig: Probably a good idea. "Core" or "Basement" could be anything.

marc-scig (commented):
It seems to work ok on my machine. Aside from ryan's suggestion about adding a prefix, it might be a good idea to try to verify the maximum file path length for a typical installation.

For example, in the open GitHub issue there's the path `C:\Users\Alex\.nuget\packages\urbanairship.ios.automation\14.4.1\content\AirshipBindings.iOS.Automation.resources\AirshipAutomation.xcframework\ios-arm64_x86_64-simulator\AirshipAutomation.framework\PrivateHeaders\UAScheduleTriggerContextTransformer+Internal.h`

With your current change this would be shortened from 261 to 241 characters. I'm not sure about whether there are longer ones though, especially with an added prefix.

khmMouna:
I just noticed that unfortunately this fix will not work with the service extension and content extension. we are going to have something like that:
**C:\Users\xxxxx\.nuget\packages\urbanairship.ios.notificationcontentextension\16.1.2\content\NotificationContentExtension.resources\AirshipNotificationContentExtensionr.xcframework\ios-arm64_x86_64-simulator\AirshipPreferenceCenter.framework\Headers\AirshipNotificationContentExtension.h** which exceed the 255 (it's about 299) unless we want to reduce the xcframework and the package name length  but I am not sure we want to that for the moment.

rlepinski:
I think we should continue with this change and plan to further reduce names later

rlepinski (approved):

oristanovic (approved):

PR #193, opened 2022-02-14
Mobile 2775 Reduce nuget path longth

khmMouna:
### What do these changes do?
Reduce the nuget package name as possible and use Airship instead of Urbanairship

Fix an issue with windows nuget long path.

Run sample app and test manually.


Airship.iOS.Basement.nuspec
```diff
@@ -15,7 +15,7 @@
 
    <files>
      <file src="src/AirshipBindings.iOS.Basement/bin/release/AirshipBindings.iOS.Basement.dll" target="lib/Xamarin.iOS10" />
-     <file src="src/AirshipBindings.iOS.Basement/bin/Release/AirshipBindings.iOS.Basement.resources/**" target="content/UABasement.resources" />
-     <file src="src/AirshipBindings.iOS.Basement/urbanairship.ios.basement.targets" target="build/urbanairship.ios.basement.targets" />
+     <file src="src/AirshipBindings.iOS.Basement/bin/Release/AirshipBindings.iOS.Basement.resources/**" target="content/ABasement.resources" />
```
rlepinski, line 18: This seems to be inside our module right? Can we just say `content/res.resources` ?

rlepinski (approved):
Looks good, wonder if we can just normalize on a resource path though

Airship.iOS.Accengage.nuspec
```diff
@@ -15,7 +15,7 @@
 
    <files>
      <file src="src/AirshipBindings.iOS.Accengage/bin/release/AirshipBindings.iOS.Accengage.dll" target="lib/Xamarin.iOS10" />
-     <file src="src/AirshipBindings.iOS.Accengage/bin/Release/AirshipBindings.iOS.Accengage.resources/**" target="content/UAAccengage.resources" />
```
rlepinski, line 18: I think we can just do `content/res` for all of them

rlepinski:
Is it possible to also rename the xcframeworks that we are consuming? AirshipPreferenceCenter.xcframework is pretty long

oristanovic (approved):

PR #198, opened 2022-05-06
Release xamarin 16.1.0

khmMouna:
### What do these changes do?
Prepare the release xamarin 16.1.0
#### Changes
- Updated Android SDK to 16.4.0
- Updated iOS SDK to 16.6.0
- Add Room runtime as a dependency on Android Core
- Add InAppMessagingDelegate method and notifyDisplayConditionsChanged on iOS

Sample app


src/AirshipBindings.NETStandard/AirshipBindings.NETStandard.iOS/Airship.cs
```diff
@@ -304,34 +304,35 @@ public void AddCustomEvent(CustomEvent customEvent)
...
                     {
-                        propertyDictionary.SetValueForKey((NSNumber)doubleProperty.value, (NSString)doubleProperty.name);
+                        propertyDictionary.SetValueForKey((NSNumber)doubleProperty.value, key);
                     }
                     else if (property is CustomEvent.Property<bool> boolProperty)
                     {
-                        propertyDictionary.SetValueForKey((NSNumber)boolProperty.value, (NSString)boolProperty.name);
+                        propertyDictionary.SetValueForKey((NSNumber)boolProperty.value, key);
                     }
                     else if (property is CustomEvent.Property<string[]> stringArrayProperty)
```
rlepinski, line 328: We can accept any JSOn value now, do we really need to iterate through?

khmMouna: I think we can factorised by casting it to NSObject directly but It should be working with string, bool and double but still need to cast it for the array:

```
 NSString key = (NSString)property.name;
 NSObject value = NSObject.FromObject(property.value);

 if (property is CustomEvent.Property<string[]> stringArrayProperty)
 {
    value = NSArray.FromObjects(stringArrayProperty.value);
    propertyDictionary.SetValueForKey(value, key);
 }
 if (value != null)
 {
     propertyDictionary.SetValueForKey(value, key);
 }
```

What do you think ?

Ulrico972 (approved):

rlepinski (approved):

PR #217, opened 2023-06-29
[MOBILE-3703] Release 17.1.0

jyaganeh:
### What do these changes do?

- Bumps to cross-platform version 17.1.0, Android SDK 16.11.1, iOS SDK 16.12.2
- Added AndroidX WebKit as a dependency of the Android Core binding

Integrating bugfixes from SDK 16.x releases.

Ran the Android and iOS sample app and verified channel registration, MC, PC, actions, IAA, etc.


rlepinski (approved):

rlepinski (approved):

SampleApp/SampleApp/MessageCenterViewController.xaml.cs
```diff
@@ -18,10 +17,22 @@ public MessageCenterViewController()
...
+            Console.WriteLine("Inbox Messages:");
+
+            foreach (Message message in messages)
+            {
+                Console.WriteLine("---");
+                Console.WriteLine(message.Title);
+                Console.WriteLine(message.Unread);
+                Console.WriteLine(message.SentDate);
+            }
+            Console.WriteLine("---");
```
Ulrico972, line 30: was this for test purpose ?

Ulrico972 (approved):

SampleApp/SampleApp/MessageCenterViewController.xaml.cs
jyaganeh: Yeah, the linker was stripping out some fields from the `Message` class when the linking mode was set to `Full`, and that was causing the message list to look blank/empty, even though there were clickable messages in the list. 🤷 

I added this so I could see what was going on and thought it might be helpful to leave, since it's in the sample code and could be helpful if anyone runs into similar issues with their linker setting.

Ulrico972: Oh right, I didn't notice 👍

PR #223, opened 2023-10-02
[MOBILE-3929] release-18.0.0 - Bump to latest SDKs

jyaganeh:
### What do these changes do?

* Bumps to iOS 17.4.0, Android 17.3.0
* Updated iOS `Airship.cs` to use notification status update constants defined on `UAPush` (also updated core api definitions to declare them)
* Updated the Sample app to use the new nupkgs
* Fixed Android tab bar so it appears at the bottom of the screen and highlights the selected item in blue
* Commented out packing of android feature flag bindings (we'll re-enable when iOS support is also available)
* Fixed a couple of minor warnings/suggestions in the Sample app

Last bits of prep for 18.0.0 release!

Tested iOS and Android sample apps and checked things out in Go.


khmMouna (approved):
Nice catch for notification status 
Minor update to fix failed build: Need to update ios extensions in Sample app

rlepinski (approved):

PR #226, opened 2024-01-18
[MOBILE-4082] Fix missing custom event properties and log level trace->verbose on iOS

jyaganeh:
### What do these changes do?

- Uncomments the line that adds custom event properties to the native custom event object
- Deprecates `UALogLevel.Trace` and adds `UALogLevel.Verbose` (with the same value, so they'll do the same thing, but `Trace` will show a warning in the IDE)

Customer reported custom events were missing properties on iOS, and we changed the LogLevel in SDK 17, but missed the rename in this repo.

* Reproduced the customer's issue using RTDS to view the payloads
* Wired up a custom event to a button on the sample app home screen and verified that the correct payloads with properties were showing up in the RTDS stream
* Migrated our single usage of `Trace` to `Verbose` and ran the sample app

### DIFF
--- a/src/AirshipBindings.iOS.Core/StructsAndEnums.cs
+++ b/src/AirshipBindings.iOS.Core/StructsAndEnums.cs
@@ -49,7 +49,9 @@ namespace UrbanAirship
 		Warn = 2,
 		Info = 3,
 		Debug = 4,
-		Trace = 5
+		[Obsolete("Use Verbose instead. Trace will be removed in a future release.")]
+		Trace = 5,
+		Verbose = 5
 	}
 
     [Native]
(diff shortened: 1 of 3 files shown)


rlepinski (approved):

src/AirshipBindings.iOS.Core/StructsAndEnums.cs
```diff
@@ -49,7 +49,9 @@ public enum UALogLevel : long
 		Warn = 2,
 		Info = 3,
 		Debug = 4,
-		Trace = 5
+		[Obsolete("Use Verbose instead. Trace will be removed in a future release.")]
+		Trace = 5,
+		Verbose = 5
```
jyaganeh, line 54: Duplicate enum values are allowed in C#, so I did this in order to avoid a breaking change. Maybe slightly funky, but we can get rid of it in the next major.

khmMouna (approved):

